PR #931, opened 2015-02-27
Update AbstractGoogleClientRequest.java

elibixby:
Adds an `execute(int)` method which automatically catches exceptions and tries exponential backoff


elibixby:
@wonderfly PTAL

wonderfly:
AbstractGoogleClientRequest does give you a way to get access to the underlying HttpRequest through its buildHttpRequest() method. Then you are free to register handlers through its setters.

https://github.com/google/google-api-java-client/blob/master/google-api-client/src/main/java/com/google/api/client/googleapis/services/AbstractGoogleClientRequest.java#L276

elibixby:
Yes, but the default flow does not involve getting the HttpRequest, executing the HttpRequest, then parsing it. All of this additional work should not be necessary in order to set the number of retries of a client request (this is not the case in other client libraries).

elibixby:
@wonderfly PTAL

PR #946, opened 2015-06-11
Fix for java.security.SignatureException: Signature length not correct

davidpaniz:
Since a few hour ago, oath integration with google is not working because a exception `java.security.SignatureException: Signature length not correct: got 256 but was expecting 128`

This solution is based on this comment http://stackoverflow.com/questions/30780407/google-oauth2-jwt-token-verification-exception
If a exception is thrown while checking keys, it returns false instead of break.

### DIFF
--- a/google-api-client/src/main/java/com/google/api/client/googleapis/auth/oauth2/GoogleIdTokenVerifier.java
+++ b/google-api-client/src/main/java/com/google/api/client/googleapis/auth/oauth2/GoogleIdTokenVerifier.java
@@ -171,8 +171,12 @@ public class GoogleIdTokenVerifier extends IdTokenVerifier {
     }
     // verify signature
     for (PublicKey publicKey : publicKeys.getPublicKeys()) {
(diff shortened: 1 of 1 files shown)


aaronharnly:
@davidpaniz if you haven't, it looks like you need to follow https://developers.google.com/api-client-library/java/google-api-java-client/contribute

davidpaniz:
Hi @aaronharnly,

I really skipped all the steps, sorry about that. This PR fixes #947, I did it in a rush to solve a production issue and thought I might be helpful to you. Maybe you can just close this PR and use the main ideia to solve the issue. In the meantime I will follow the steps to contribute to the project.

ejona86:
We'll be putting a try-catch around google-http-client's [SecurityUtils.verify](https://github.com/google/google-http-java-client/blob/dev/google-http-client/src/main/java/com/google/api/client/util/SecurityUtils.java#L164) to fix the issue in the future.

PR #952, opened 2015-07-17
Use lattest version of library

neshkeev:
Use the lattest version of the google-api-client library


ejona86:
Great idea! :-) Thanks!

PR #953, opened 2015-07-23
Adds support for retrieving credentials in the Cloud Shell environment.

wilford:


anthmgoogle:
LGTM, although you will need to squash the commits to a single one before merging.

I don't have permission to merge. We can track that down separately.

google-api-client/pom.xml
```diff
@@ -127,10 +128,20 @@
...
-   </dependency>
+    </dependency>
     <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava-jdk5</artifactId>
     </dependency>
+    <dependency>
+      <groupId>javax.json</groupId>
+      <artifactId>javax.json-api</artifactId>
+      <version>1.0</version>
```
ejona86: The dependencies with versions normally go in the parent's pom.xml, and then the version isn't specified here. That guarantees that all the projects use the same version of dependencies.

ejona86:
@anthmgoogle, I've got write access, I can help you out.

How do we feel about adding a JSON library dependency to the core api-client? It seems we have gone to great lengths to avoid that, I wouldn't expect us to start now.

google-api-client/pom.xml
```diff
@@ -127,10 +128,20 @@
...
       <scope>provided</scope>
       <version>1.6</version>
-   </dependency>
+    </dependency>
     <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava-jdk5</artifactId>
     </dependency>
+    <dependency>
+      <groupId>javax.json</groupId>
```
anthmgoogle: We should specifically avoid these javax.json dependencies, since this version of the API must run on a broad set of platforms that can have restrictions about this. Make use of the existing com.google.api.client.json instead.

anthmgoogle:
Undo my LGTM. @ejona86 yes, the new json dependency should be avoided here.

google-api-client/pom.xml
wilford: Removed.

google-api-client/pom.xml
wilford: Removed.

wilford:
Sorry, really messed up my local client so I closed this one and created a
separate pull request in
https://github.com/google/google-api-java-client/pull/955 that should
address your comments.

PR #955, opened 2015-07-29
Adds support for retrieving credentials in the Cloud Shell environment.

wilford:


anthmgoogle:
LGTM.

google-api-client/src/main/java/com/google/api/client/googleapis/auth/oauth2/CloudShellCredential.java
```diff
@@ -0,0 +1,97 @@
...
+    try {    
+      PrintWriter out =
+        new PrintWriter(socket.getOutputStream(), true);
+      out.println(GET_AUTH_TOKEN_REQUEST);
+
+      BufferedReader input =
+          new BufferedReader(new InputStreamReader(socket.getInputStream()));
+      String ignoredSizeLine = input.readLine();
+      Collection<String> messageArray = GsonFactory.getDefaultInstance()
+	.createJsonParser(input)
```
ejona86: Could we please not use tabs for indentation?

google-api-client/pom.xml
```diff
@@ -110,7 +111,6 @@
     <dependency>
       <groupId>com.google.http-client</groupId>
       <artifactId>google-http-client-gson</artifactId>
-      <scope>test</scope>
```
ejona86: This still adds a new dependency on gson. I think you may want to use the JsonFactory interface from google-http-java-client.

google-api-client/src/main/java/com/google/api/client/googleapis/auth/oauth2/CloudShellCredential.java
wilford: Done.

google-api-client/pom.xml
wilford: JsonFactory is an interface though, so which implementing class should I use?  It seems that any class I use adds a dependency?

wilford:
Which implementing class of JsonFactory should I use?  It seems that all of them would add another dependancy?

google-api-client/pom.xml
ejona86: A cursory look around seems like code just expects it to be passed in as a parameter. There does exist `Util.getDefaultJsonFactory()`, and it seems there is already a dependency on Jackson. However, the only place it is used is in `GoogleCredential`.

https://github.com/google/google-api-java-client/blob/57fe35766cbba0a0d5a9a296be81468d730a29f8/google-api-client/src/main/java/com/google/api/client/googleapis/compute/ComputeCredential.java#L70
https://github.com/google/google-api-java-client/blob/57fe35766cbba0a0d5a9a296be81468d730a29f8/google-api-client/src/main/java/com/google/api/client/googleapis/auth/oauth2/GooglePublicKeysManager.java#L93
https://github.com/google/google-api-java-client/blob/57fe35766cbba0a0d5a9a296be81468d730a29f8/google-api-client/src/main/java/com/google/api/client/googleapis/json/GoogleJsonResponseException.java#L92
https://github.com/google/google-api-java-client/blob/57fe35766cbba0a0d5a9a296be81468d730a29f8/google-api-client/src/main/java/com/google/api/client/googleapis/auth/oauth2/GoogleClientSecrets.java#L190
https://github.com/google/google-api-java-client/blob/57fe35766cbba0a0d5a9a296be81468d730a29f8/google-api-client/src/main/java/com/google/api/client/googleapis/auth/oauth2/GoogleAuthorizationCodeTokenRequest.java#L90
https://github.com/google/google-api-java-client/blob/57fe35766cbba0a0d5a9a296be81468d730a29f8/google-api-client/src/main/java/com/google/api/client/googleapis/auth/oauth2/GoogleRefreshTokenRequest.java#L84

google-api-client/src/main/java/com/google/api/client/googleapis/auth/oauth2/CloudShellCredential.java
```diff
@@ -0,0 +1,102 @@
...
+    try {    
+      PrintWriter out =
+        new PrintWriter(socket.getOutputStream(), true);
+      out.println(GET_AUTH_TOKEN_REQUEST);
+
+      BufferedReader input =
+          new BufferedReader(new InputStreamReader(socket.getInputStream()));
+      String ignoredSizeLine = input.readLine();
+      Collection<String> messageArray = GsonFactory.getDefaultInstance()
+<<<<<<< HEAD
```
anthmgoogle: Incomplete merge?

google-api-client/pom.xml
anthmgoogle: Yes, you should take the JsonFactory as a parameter to CloudShellCredential, and use the one that getApplicationDefault has had passed in. If you want to have a default value on CloudShellCredential for cases where people use it directly, use Util.getDefaultJsonFactory().

google-api-client/src/main/java/com/google/api/client/googleapis/auth/oauth2/CloudShellCredential.java
wilford: Done.

google-api-client/pom.xml
wilford: Done.

wilford:
Fixed deps issues.

ejona86:
This LGTM. @anthmgoogle still look good to you?

@wilford if you're ready and @anthmgoogle LGTMs (because I consider mine more of OWNERS approval), then I'll go ahead and merge it.

wilford:
I'm ready whenever @anthmgoogle <https://github.com/anthmgoogle>  is.
Thanks!

anthmgoogle:
LGTM. Thanks. We should be ready to merge now.

ejona86:
@wilford, checkstyle fails. Please run `mvn verify` and fix the style failures.

wilford:
Ah, my bad.  Fixed checkstyle errors.

ejona86:
Cherry-picked as 3d866eb7

PR #966, opened 2015-08-28
Possible implementation of resumable uploads.

geoffgibbs:
I have used this variation of the MediaHttpUploader to allow for resumable uploads to be paused and restarted in a project of my own.

### DIFF
--- a/google-api-client/src/main/java/com/google/api/client/googleapis/media/MediaHttpUploader.java
+++ b/google-api-client/src/main/java/com/google/api/client/googleapis/media/MediaHttpUploader.java
@@ -412,7 +412,8 @@ public final class MediaHttpUploader {
    * @return the response
    * @throws IOException if there is a communication error
    */
-  private HttpResponse resumableUploadInitiation(GenericUrl resumableRequestUrl) throws IOException {
+  private HttpResponse resumableUploadInitiation(GenericUrl resumableRequestUrl)
+      throws IOException {
     updateStateAndNotifyListener(UploadState.RESUME_INITIATION_STARTED);
     HttpRequest request =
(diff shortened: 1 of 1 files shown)


IvanRF:
Could you include an example on how you use it?

I guess that you set a listener `request.getMediaHttpUploader().setProgressListener(MediaHttpUploaderProgressListener);` and on every `UploadState.MEDIA_IN_PROGRESS` you store the current state with the new `getResumableUploadUrl()`.
Now, to continue with a previous upload I guess you use `continueResumableUpload(String resumableUploadUrl)` on a new request.

I think, instead of having two different functions calling `doResumableUpload(String)`, wouldn't be better to have a `setResumableUploadUrl(String)` which modifies current variables in order  to always use the current `resumableUpload(GenericUrl)`. This way, after you create a request you can set the resumable URL if exists, and continue with the execution.

In my case, I use exponential backoff for all executions on `AbstractGoogleClientRequest`. Having to call `continueResumableUpload(String)` will break that abstraction, that's why I think a `setResumableUploadUrl(String)` would be better.

geoffgibbs:
That is how I'm using it in my code. But, to be honest, that was used to fix a specific problem at a specific time and I haven't had the need to go back to it before now.

I'll take another look and see what I can do.

geoffgibbs:
I signed it?!

chingor13:
Resumable uploads is available.

IvanRF:
@chingor13 this is a related issue: https://github.com/googleapis/google-api-java-client/issues/995
If that issue is solved now, could you post a link to the new code? thanks!

kbboss:
@chingor13  yep,how do you did it? post it,if you sloved it

dknchris:
@geoffgibbs Can you please share a gist of how to actually use the resumable upload?

PR #968, opened 2015-10-12
google-api-java-client-xml: Depend on the correct version of google-http-java-client-xml

wonderfly:


ejona86:
@wonderfly, it looks like #969 was a duplicate of this one, but I already merged #969, so should we close this?

wonderfly:
yes, please

PR #976, opened 2015-11-01
Updated default issuer

Gameleon12:
Updated the default issuer to `https://accounts.google.com` to match the
updated issuer value contained in new Google id tokens.

The original `accounts.google.com` value caused new tokens to fail verification when not setting the issuer manually using `setIssuer("https://accounts.google.com")`


ejona86:
This isn't enough, since it will break any google id tokens using the old endpoint. We will actually need to support both strings at the same time. I already have a change made, I just need to bring it into this repo.

ejona86:
This problem was fixed by #982 instead, which allows both the old and the new issuer to function.

PR #985, opened 2015-11-15
Change lastBytePos to long in MediaHttpDownloader.setContentRange

### LINKED ISSUE #937: MediaHttpDownloader.setContentRange uses int lastBytePos
firstBytePos is a long, but the lastBytePos is an int, which truncates the offset of a range you can specify.

mziccard:
This PR tries to fix #937.

This PR breaks `clirr` check with the following error:

> Parameter 2 of 'public com.google.api.client.googleapis.media.MediaHttpDownloader setContentRange(long, int)' has changed its type to long

I don't see any major issue with this as `MediaHttpDownloader.lastBytePos` field is already a `long`, but correct me if I'm wrong.

### DIFF
--- a/google-api-client/src/main/java/com/google/api/client/googleapis/media/MediaHttpDownloader.java
+++ b/google-api-client/src/main/java/com/google/api/client/googleapis/media/MediaHttpDownloader.java
@@ -304,7 +304,7 @@ public final class MediaHttpDownloader {
    * @param lastBytePos The last byte position in the content range string.
    * @since 1.13
    */
-  public MediaHttpDownloader setContentRange(long firstBytePos, int lastBytePos) {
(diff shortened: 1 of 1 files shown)


ahmadyousaf:
As if this functions receives an int, assigning it to long would still have int data. By definition, in java int can hold maximum of 2^31 - 1
If we have a file just larger than 2gb. we cannot specify lastBytePos if it's at the end of the file. On the same note, if you could also update the javadocs for this function as it says: "Sets the content range of the next download request. Eg: bytes=firstBytePos-lastBytePos."

It should rather be "... Eg: bytes=lastBytePos-firstBytePos."

jamie-everlaw:
Any reason this hasn't been merged?

PR #986, opened 2015-11-20
Allow PATCH as a valid initial method for uploads

sqrrrl:


ejona86:
This LGTM. In a separate thread it sounded like you were running into some issues when trying to use it. Just ping here or on that separate thread once you feel comfortable with the change and I'll merge it.

Thanks for adding the tests.

sqrrrl:
I verified it works with direct/multipart uploads. There is a separate issue with resumable uploads, but that appears to be an issue on the backend and is reproducible with other client libraries too.

ihommani:
Hi, 

@ejona86 , is there any repo with the release candidate containing this fix available ?
Thx

ejona86:
@ihommani, this repo contains the fix. Do you mean Maven repo? No, but I can make a snapshot and I will be doing a new release soon.

ihommani:
@ejona86 , yes a maven repo. But it's ok for now, I used our own nexus server to host the snapshot.  Waiting for the new release though :)

Thx

PR #987, opened 2015-11-23
Fix of server google server exception. PublicKeys are still considered…

tunkl:
… valid if communication error occurred.

### DIFF
--- a/google-api-client/src/main/java/com/google/api/client/googleapis/auth/oauth2/GooglePublicKeysManager.java
+++ b/google-api-client/src/main/java/com/google/api/client/googleapis/auth/oauth2/GooglePublicKeysManager.java
@@ -21,7 +21,12 @@ import com.google.api.client.http.HttpTransport;
 import com.google.api.client.json.JsonFactory;
 import com.google.api.client.json.JsonParser;
 import com.google.api.client.json.JsonToken;
-import com.google.api.client.util.*;
+import com.google.api.client.util.Beta;
+import com.google.api.client.util.Clock;
+import com.google.api.client.util.Lists;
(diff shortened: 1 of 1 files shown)


tunkl:
I signed it!

tunkl:
Integration test failed because of maven problem. Is there some way how to run it again?

ejona86:
@tunkl, I was able to restart the Travis build and they completed successfully.

google-api-client/src/main/java/com/google/api/client/googleapis/auth/oauth2/GooglePublicKeysManager.java
```diff
@@ -21,11 +21,7 @@
 import com.google.api.client.json.JsonFactory;
 import com.google.api.client.json.JsonParser;
 import com.google.api.client.json.JsonToken;
-import com.google.api.client.util.Beta;
-import com.google.api.client.util.Clock;
-import com.google.api.client.util.Preconditions;
-import com.google.api.client.util.SecurityUtils;
-import com.google.api.client.util.StringUtils;
+import com.google.api.client.util.*;
```
ejona86: Although checkstyle didn't complain and we are lacking a CONTRIBUTING document to inform you of our style, we [don't use wildcard imports](https://google.github.io/styleguide/javaguide.html#s3.3.1-wildcard-imports).

ejona86:
I'm a little concerned that swallowing the exception will defeat exponential backoff of callers. But since we don't provide built-in exponential backoff and the API doesn't help callers coordinate (if one is waiting due to backoff, it won't prevent another caller from causing another refresh), I'm thinking this can be fine.

PR #992, opened 2015-12-28
Fixes issue where Cloud Shell credential would fail to parse integers…

wilford:
… returned in the JSON response.


ejona86:
@wilford, thanks.

PR #998, opened 2016-02-05
Remove oraclejdk7 from travis build

ejona86:
oraclejdk7 is no longer supported by Oracle and openjdk7 is pretty
similar.


nmittler:
LGTM

PR #1002, opened 2016-03-14
Moe sync

ejona86:


ejona86:
@nmittler, only the "Avoid diamond operator" is an original commit. I did manually resolve some merge conflicts though.

google-api-client/src/main/java/com/google/api/client/googleapis/auth/oauth2/DefaultCredentialProvider.java
```diff
@@ -216,6 +161,104 @@ String getProperty(String property, String def) {
...
+      return fileExists(wellKnownFileLocation);
+    } catch (AccessControlException expected) {
+      // Exception querying file system is expected on App-Engine
+      return false;
+    }
+  }
+
+  private GoogleCredential getCredentialUsingWellKnownFile(
+      HttpTransport transport, JsonFactory jsonFactory) throws IOException {
+    // Then try the well-known file
```
nmittler: nit: remove this comment? I'm assuming this code was broken out from a larger if-then-else?

google-api-client/src/main/java/com/google/api/client/googleapis/auth/oauth2/OAuth2Utils.java
```diff
@@ -21,18 +21,34 @@
...
 import java.util.Collection;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 /**
  * Utilities used by the com.google.api.client.googleapis.auth.oauth2 namespace.
- *
  */
-class OAuth2Utils {
+public class OAuth2Utils {
```
nmittler: Are you expecting the application to use this directly? Should it be marked as Beta or something indicating API-stability guarantees?

google-api-client/src/main/java/com/google/api/client/googleapis/auth/oauth2/OAuth2Utils.java
```diff
@@ -53,20 +69,46 @@ static boolean headersContainValue(HttpHeaders headers, String headerName, Strin
...
+
+    GenericUrl tokenUrl = new GenericUrl(getMetadataServerUrl(environment));
+    for (int i = 1; i <= MAX_COMPUTE_PING_TRIES; ++i) {
+      try {
+        HttpRequest request = transport.createRequestFactory().buildGetRequest(tokenUrl);
+        request.setConnectTimeout(COMPUTE_PING_CONNECTION_TIMEOUT_MS);
+        HttpResponse response = request.execute();
+        try {
+          HttpHeaders headers = response.getHeaders();
+          return headersContainValue(headers, "Metadata-Flavor", "Google");
```
nmittler: I've seen this header key/value before (above) ... should we make them constants?

nmittler:
@ejona86 Looks good! One general comment that I think a lot of the string values you're using throughout could be turned into constants.

google-api-client/src/main/java/com/google/api/client/googleapis/auth/oauth2/OAuth2Utils.java
ejona86: In this particular case I don't think I can give them better names as constants, so I think converting them to constants would be unhelpful. If you have good ideas for names, then maybe I could change my mind.

ejona86:
@nmittler, PTAL

google-api-client/src/main/java/com/google/api/client/googleapis/auth/oauth2/OAuth2Utils.java
nmittler: It's more that it reduces a risk of a type-o.  Better to be certain that we're consistent ... even if consistently wrong :).

ejona86: Then please provide useful constant names; I don't have any. The ones I've thought of make the code unreadable, like `GOOGLE` and `MAGIC_VALUE` and `GOOGLE_FLAVOR`. I dislike the constants when they reduce readability and encourage bugs because they added a harmful layer of abstraction. When the new names are useful and nicely abstracted then it improves the readability and can reduce the likelihood of bugs.

nmittler: I disagree that it's a harmful layer or that it encourages bugs ... but let's not get into a religious debate here.

I think your names are probably fine ...  something like `COMPUTE_HEADER_KEY` and `COMPUTE_HEADER_VALUE` would also be fine.

But I'm fine with whatever you want to do ... your call.

nmittler:
LGTM

PR #1023, opened 2016-09-07
Don't throw NPE on GoogleCredential.Builder when scopes not set

CydeWeys:
The list of account scopes in the GoogleCredential.Builder starts off
uninitialized (i.e. null), meaning that calling the .build() method
throws an NPE unless setServiceAccountScopes() is called.  There are no
other fields in the builder that are required to be called, and this
doesn't seem to be required intentionally, so make the Builder resilient
to it being null and have it correctly interpret that as an empty list
of scopes.


CydeWeys:
Is Travis's setup broken?  It's throwing an error about being unable to install google_play_services, which seems completely unrelated to my changes.  Note that all tests pass locally.

mattwhisenhunt:
Already merged: https://github.com/google/google-api-java-client/commit/44a6f91efd41c43c2a51d7c57f2664dbc3eaeb0e

PR #1029, opened 2016-11-16
Update gradle/maven examples to 1.22.0

davegwatson:


mattwhisenhunt:
Thanks for the fix, Dave. We missed out 1.22 window, but this PR will live on in spirit: #1071.

PR #1034, opened 2017-02-05
add OSGI metadata

elharo:
@briandealwis @chanseokoh @garrettjonesgoogle

Note especially that this PR moves the jar goal from the compile phase to the default package phase. Was there a reason this was in the compile phase?

### DIFF
--- a/google-api-client/pom.xml
+++ b/google-api-client/pom.xml
@@ -36,19 +36,41 @@
         <configuration>
           <archive>
             <manifest>
-              <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
+<addDefaultImplementationEntries>true</addDefaultImplementationEntries>
             </manifest>
+             <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
           </archive>
         </configuration>
         <executions>
           <execution>
             <id>jar</id>
-            <phase>compile</phase>
+            <phase>package</phase>
             <goals>
               <goal>jar</goal>
             </goals>
           </execution>
         </executions>
+      </plugin>
+      <plugin>   
+        <groupId>org.apache.felix</groupId>
+        <artifactId>maven-bundle-plugin</artifactId>
+        <version>2.5.4</version>
+        <executions>
+          <execution>
+            <id>bundle-manifest</id>
+            <phase>process-classes</phase>
+            <goals>    
+              <goal>manifest</goal>
+            </goals>   
+          </execution>
+        </executions>
+        <configuration>
+          <instructions>
+            <Bundle-DocURL>https://developers.google.com/api-client-library/java/</Bundle-DocURL>
+            <Bundle-Description>The Google API Client Library for Java provides functionality common to all Google APIs; for example HTTP transport, error handling, authentication, JSON parsing, media download/upload, and batching.</Bundle-Description>
+            <Bundle-SymbolicName>com.google.api.client.googleapis</Bundle-SymbolicName>
+          </instructions>
+        </configuration>
       </plugin>
       <plugin>
         <artifactId>maven-source-plugin</artifactId>
--- a/google-api-client/pom.xml
+++ b/google-api-client/pom.xml
@@ -41,15 +41,6 @@
              <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
           </archive>
         </configuration>
-        <executions>
-          <execution>
-            <id>jar</id>
-            <phase>package</phase>
-            <goals>
-              <goal>jar</goal>
-            </goals>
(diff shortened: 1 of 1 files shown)


briandealwis:
The resulting jar looks good to me.  Would it be worth doing some of the other libraries too?

elharo:
Maybe, though I want to straighten out the travis failure first. Which other of these jars do we use?

briandealwis:
Hmm, maybe we don't use anything else.  `com.google.cloud.tools.eclipse.login` includes a `lib/google-api-services-oauth2-v2-rev114-1.22.0.jar` that that appears to come from `com.google.apis:google-api-services-oauth2`, a different repo.

google-api-client/pom.xml
```diff
@@ -36,20 +36,42 @@
...
+<addDefaultImplementationEntries>true</addDefaultImplementationEntries>
             </manifest>
+             <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
           </archive>
         </configuration>
         <executions>
           <execution>
             <id>jar</id>
-            <phase>compile</phase>
+            <phase>package</phase>
```
garrettjonesgoogle: Just remove this executions section like was done here:

https://github.com/google/google-oauth-java-client/commit/57fd88b34bea69005c0a3bf0ad547a7a3acb2005

Also, like in the above commit, remove the `phase` from maven-source-plugin below.

google-api-client/pom.xml
```diff
@@ -36,20 +36,42 @@
...
+            <goals>    
+              <goal>manifest</goal>
+            </goals>   
+          </execution>
+        </executions>
+        <configuration>
+          <instructions>
+            <Bundle-DocURL>https://developers.google.com/api-client-library/java/</Bundle-DocURL>
+            <Bundle-Description>The Google API Client Library for Java provides functionality common to all Google APIs; for example HTTP transport, error handling, authentication, JSON parsing, media download/upload, and batching.</Bundle-Description>
+            <Bundle-SymbolicName>com.google.api.client.googleapis</Bundle-SymbolicName>
```
garrettjonesgoogle: It looks like you're using a different set of instructions from https://github.com/google/google-oauth-java-client/commit/57fd88b34bea69005c0a3bf0ad547a7a3acb2005 - what are the reasons?

google-api-client/pom.xml
```diff
@@ -36,20 +36,42 @@
         <configuration>
           <archive>
             <manifest>
-              <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
+<addDefaultImplementationEntries>true</addDefaultImplementationEntries>
```
chanseokoh: Indentation went off here.

google-api-client/pom.xml
elharo: done

google-api-client/pom.xml
elharo: This one doesn't seem to need to provide a specific Java 6 bundle. I should probably add a Bundle-DocURL and Bundle-Description to that one.

google-api-client/pom.xml
garrettjonesgoogle: Looks like it still needs to be reverted.

google-api-client/pom.xml
```diff
@@ -36,26 +36,38 @@
...
-            <id>jar</id>
-            <phase>compile</phase>
-            <goals>
-              <goal>jar</goal>
-            </goals>
+            <id>bundle-manifest</id>
+            <phase>process-classes</phase>
+            <goals>    
+              <goal>manifest</goal>
+            </goals>
```
garrettjonesgoogle: Remove trailing whitespace

elharo:
Fails with


The command "mvn install:install-file -Dfile=$ANDROID_HOME/extras/google/google_play_services/libproject/google-play-services_lib/libs/google-play-services.jar -DgroupId=com.google.android.google-play-services -DartifactId=google-play-services -Dversion=1 -Dpackaging=jar" failed and exited with 1 during .

That might be related to these changes.

briandealwis:
It looked like other PRs fail with three same issue.

chanseokoh:
I did some investigation. The failure has nothing to do with this PR; other previous PRs had this failure too. It's certain that builds of this repo have been broken for a while.

The problem is that the latest Android SDK doesn't have `<android-sdk-home>/extras/google/google_play_services/libproject/google-play-services_lib/`, from which the current build tries to get `google-play-services.jar`. (There are many internal bugs related to this: 29869239, 28866296, etc.) One of the sub-modules, `google-api-client-android`, wants to use the jar.

I had a chat with @loosebazooka, but it's still not clear how we can pull in this dependency.

PR #1036, opened 2017-03-01
added impersionation overloads for building from stream

### LINKED ISSUE #1007: Service account: Cannot set user to impersonate using GoogleCredential fromStream
I'm trying to use a service account to automatically manage/create groups using the Directory API. When I try to use a service account with the credentials in the form of JSON and try to use the GoogleCredential.fromStream(...) call, there does not seem to be a way to set the user, and at least for the Admin API I cannot get calls to work without impersonating a user as I always get Forbidden (403) responses. 

Using the GoogleCredential Builder seems to work fine with the PKCS12 formatted key as long as I set the user, but your docs and console seem to recommend to use the JSON formatted credential file. 

Have I missed something, or is this functionality not available without using the builder?

woodyspawnshop:
fixes #1007


woodyspawnshop:
I signed it!

woodyspawnshop:
because your [CI is broke](https://travis-ci.org/google/google-api-java-client/jobs/206775245#L615-L624)

woodyspawnshop:
Until the grpc libs show up can you pretty please accept this pull request? Its strait forward and it solves a glaring gap in the client lib that nearly everyone experiences because the default behavior in cloud console is to supply `.json`, not `.p12`. I feel this qualifies as a necessary bug\essential feature.

hgschmie:
I wrote a different implementation to solve this problem in #1079. Please take a look.

nbali:
is there any reason this hasn't been merged?

PR #1048, opened 2017-04-24
Possible Fix for AppIdentityCredentialWrapper.getAccessToken() Method

hiranya911:
This method currently always returns null in App Engine environment, which is a major problem when working with application default credentials. Also the associated `refreshToken()` method always returns false. This fix enables users to call `refreshToken()` and `getAccessToken()` on application default credentials in the App Engine environment. 

I'm not very familiar with this codebase, and hence it may not be complete or even fully accurate. Feedback appreciated.


saicheems:
+@anthmgoogle

hiranya911:
Thanks for approving the patch. What's the process for getting this merged into the codebase? And what does your release schedule look like for the next few months?

PR #1049, opened 2017-04-27
Latest version is now 1.22.0

rcleveng:
Update README.md to use 1.22.0 as the latest version.


rcleveng:
It doesn't seem the TravisCI has been working correctly for ~11 months now.

mattwhisenhunt:
Thanks Rob. We missed our 1.22 window completely, but this PR will live on in spirit: #1071

PR #1079, opened 2017-12-08
adds methods to create delegated credentials and builder

hgschmie:
adds a toBuilder() method which creates a mutable version of the credential
object preloaded with all values from the credential.

adds a createDelegated(String user) method which allows creation of
a credential with correct user delegation from an existing credential

### DIFF
--- a/google-api-client/src/main/java/com/google/api/client/googleapis/auth/oauth2/GoogleCredential.java
+++ b/google-api-client/src/main/java/com/google/api/client/googleapis/auth/oauth2/GoogleCredential.java
@@ -483,18 +483,47 @@ public class GoogleCredential extends Credential {
     if (serviceAccountPrivateKey == null) {
       return this;
     }
-    return new GoogleCredential.Builder()
+    return toBuilder()
+        .setServiceAccountScopes(scopes)
+        .build();
+  }
+
+  /**
+   * {@link Beta} <br/>
+   * For service accounts that need to delegate to a specific user, create a
+   * copy of the credential with the specified user
+   */
+  @Beta
+  public GoogleCredential createDelegated(String user) {
+    if (serviceAccountPrivateKey == null) {
+      return this;
+    }
+    return toBuilder()
+        .setServiceAccountUser(user)
+        .build();
+  }
+
+  /**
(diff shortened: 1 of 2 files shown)


hgschmie:
the ci build seems to be broken (and fails with some android error message).

nbali:
is there any reason this hasn't been merged?

woodyspawnshop:
This would require more elaborate changes from folks to take advantage of. What I did in #1036 was take an existing flow with an existing function with existing overloads and simply added one overload to allow supplying the impersonation user. In order for folks to take advantage of that they simply pass an additional parameter. IMO both of these are probably helpful if they don't conflict.

nbali:
TBH I prefer this one, but it's ridiculous that this comes up literally at every project I work with and had to work around it... so it annoyed me to the level that I wanted create a contribution about it. You can imagine my surprise when there are multiple, but they are blatantly ignored for so long.

nbali:
@jadekler ping

jadekler:
This client library is supported but in maintenance mode only, unfortunately. We are mostly in a bug-fixing, security-patching mode. Furthermore, this library is currently frozen while we clean up some of the internals and prepare for a release (hopefully making future releases easier). I'll keep this PR open and we'll circle back to this when we're done.

nbali:
@jadekler the bothering lack of any ETA means this could take a very long time?

nbali:
"we'll circle back to this when we're done." @jadekler @chingor13 any estimation when might that happen?

chingor13:
This feature is currently available in the newer google-auth-library-java (com.google.auth:google-auth-library-oauth2-http). It's weird that there are 2 implementations for service accounts (and application default credentials) and we'd like to consolidate into 1 if possible.

The `GoogleCredentials` provided from the google-auth-library-java will work the the google-http-java-client (the underlying transport), but there's not great documentation on how to do it. It's *likely* that we'll create a guide for using those credentials and deprecated the ones here, rather than adding more features to the implementation here. This is only in the proposal stage and we're thinking about what changes need to be made to get these libraries where we want them.

chingor13:
Verified that these commits rebased pass CI.

PR #1088, opened 2018-01-16
Update .travis.yml to use new defaults.

mattwhisenhunt:
This should fix Travis build errors. @garrettjonesgoogle PTAL


garrettjonesgoogle:
It looks like it is still failing (an issue with google-play-services:jar)

mattwhisenhunt:
@garrettjonesgoogle PTAL

This seems a bit extreme but the android module doesn't have any tests for Travis to run anyways.
The only difference in the travis.pom is:
```
-    <module>google-api-client-android</module>
-    <module>google-api-client-assembly</module>
```

.travis.pom.xml
```diff
@@ -0,0 +1,608 @@
...
+    <project.jackson-core2.version>2.9.2</project.jackson-core2.version>
+    <project.protobuf-java.version>2.6.1</project.protobuf-java.version>
+    <project.guava.version>19.0</project.guava.version>
+    <project.appengine.version>1.7.7</project.appengine.version>
+    <project.xpp3.version>1.1.4c</project.xpp3.version>
+    <project.commons-logging.version>1.1.1</project.commons-logging.version>
+    <project.httpclient.version>4.5.3</project.httpclient.version>
+    <project.httpcore.version>4.0.1</project.httpcore.version>
+    <project.jdo2-api.version>2.3-eb</project.jdo2-api.version>
+    <project.transaction-api.version>1.1</project.transaction-api.version>
```
garrettjonesgoogle: This looks like a maintenance nightmare - we have to make sure versions match in both the main pom and this travis pom. What was the reason for a separate pom?

Also, have you considered trying CircleCI instead?

mattwhisenhunt: Good point. 

CircleCI seems awesome. I wonder if there wouldn't be the same problem. Newer versions of the Android SDK hide that Play Services jar inside an .aar. The best approach would probably be to make testing google-api-java-client-samples part of the integration and build with Gradle.

mattwhisenhunt:
@garrettjonesgoogle PTAL

.travis.yml
```diff
@@ -1,20 +1,9 @@
...
-  - mvn install:install-file
-     -Dfile=$ANDROID_HOME/extras/google/google_play_services/libproject/google-play-services_lib/libs/google-play-services.jar
-     -DgroupId=com.google.android.google-play-services -DartifactId=google-play-services
-     -Dversion=1 -Dpackaging=jar
-install:
-  - mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
-script:
-  - mvn test -B
+  - sed -i 's/<module>google-api-client-android<\/module>//'  pom.xml
+  - sed -i 's/<module>google-api-client-assembly<\/module>//' pom.xml
```
garrettjonesgoogle: lol that does it...

Issue #1083, opened 2017-12-15
DefaultCredentialProvider should detect GAE / std using EnvVars


lesv:
google-api-client/src/main/java/com/google/api/client/googleapis/auth/oauth2/DefaultCredentialProvider.java

PR #1089, opened 2018-01-17
Check environment variables for appengine

mattwhisenhunt:
@lesv PTAL

### DIFF
--- a/google-api-client/src/main/java/com/google/api/client/googleapis/auth/oauth2/DefaultCredentialProvider.java
+++ b/google-api-client/src/main/java/com/google/api/client/googleapis/auth/oauth2/DefaultCredentialProvider.java
@@ -29,9 +29,7 @@ import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.lang.reflect.Constructor;
-import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
-import java.lang.reflect.Method;
 import java.security.AccessControlException;
 import java.util.Locale;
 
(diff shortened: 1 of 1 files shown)


lesv:
Let me make sure this is true on Java7 and then I'll ok.

lesv:
The change you've made will identify GAE/STD (java 7, Java 8) and GAE Flex (compat).   I still need to check non-compat, I think it might which will be a change in behavior.

lesv:
Java 8 on GAE Standard:


DATACENTER | us10
-- | --
GAE_DEPLOYMENT_ID | 402856547336086879
GAE_APPLICATION | s~XXX-003
GOOGLE_CLOUD_PROJECT | XXX-003
GAE_SERVICE | XXXX
GAE_ENV | standard
GAE_INSTANCE | 00c61b117c4997505a125e9a7aa139bffd4f3a1f32420029bec3c86d6239c3c935b3e89bd30a4a
USER | appengine
TMPDIR | /tmp
GCLOUD_PROJECT | XXX-003
GAE_VERSION | XXXX:20170722t140621
GAE_RUNTIME | java8

And Java 7 has:

GAE_ENV | standard
-- | --
GAE_RUNTIME | java7

Oops - I gave bad advice above.  First check for GAE_ENV or GAE_RUNTIME (that's GAE standard), and then check for GAE_VM (GAE Flex Compat).  If it's a normal Flex runtime, you probably want to use the normal GCP stuff.

lesv:
I've checked w/ Eng and my final comments above appear to be correct.

mattwhisenhunt:
So if we are running on Flex/Compat, runningOnAppEngine() should return false? Is that right?

lesv:
Felx/Compat RunningOnAppEngine() probably should be TRUE for most services.
Flex/non-compat - should be false

mattwhisenhunt:
@lesv PTAL

google-api-client/src/main/java/com/google/api/client/googleapis/auth/oauth2/DefaultCredentialProvider.java
```diff
@@ -259,38 +258,18 @@ private GoogleCredential getCredentialUsingWellKnownFile(
...
 
   private boolean runningOnAppEngine() {
-    Class<?> systemPropertyClass = null;
-    try {
-      systemPropertyClass = forName("com.google.appengine.api.utils.SystemProperty");
-    } catch (ClassNotFoundException expected) {
-      // SystemProperty will always be present on App Engine.
-      return false;
+    Map<String, String> env = System.getenv();
+    if (env.containsKey("GAE_ENV") && env.get("GAE_VM").equals("standard")) {
```
lesv: I don't think GAE_VM ever equals "standard"

lesv: I think you meant to check `GAE_ENV`, not `GAE_VM`

lesv:
Also, you might wish to see why TRAVIS is failing.

google-api-client/src/main/java/com/google/api/client/googleapis/auth/oauth2/DefaultCredentialProvider.java
mattwhisenhunt: Indeed, copy and paste error… Thanks

mattwhisenhunt:
Fixes #1083

mattwhisenhunt:
@lesv PTAL

lesv:
Travis is unhappy that @mattwhisenhunt is not a member of committers.  I'm going to merge.

PR #1087, opened 2018-01-16
Default to "batch" if setBatchPath is never called.

mattwhisenhunt:
fixes #1073 

@garrettjonesgoogle PTAL

Issue #1073, opened 2017-10-26
Issues submitting Dataflow jobs with v1.23.0


nicktrav:
Porting this over from https://github.com/GoogleCloudPlatform/google-cloud-java/issues/2453.

We're running into some authentication issues on the latest version (1.23.0). Here's what we see when trying to interface with GCS via some Beam / Dataflow jobs, for example:

```
Exception in thread "main" java.lang.RuntimeException: Error while staging packages
	at org.apache.beam.runners.dataflow.util.PackageUtil.stageClasspathElements(PackageUtil.java:322)
	at org.apache.beam.runners.dataflow.util.PackageUtil.stageClasspathElements(PackageUtil.java:263)
	at org.apache.beam.runners.dataflow.util.GcsStager.stageFiles(GcsStager.java:65)
	at org.apache.beam.runners.dataflow.DataflowRunner.run(DataflowRunner.java:503)
	at org.apache.beam.runners.dataflow.DataflowRunner.run(DataflowRunner.java:153)
	at org.apache.beam.sdk.Pipeline.run(Pipeline.java:295)
	at org.apache.beam.sdk.Pipeline.run(Pipeline.java:281)
Caused by: java.lang.RuntimeException: Could not stage classpath element: /Users/nickt/Development/java/dist/realtime.jar
	at org.apache.beam.runners.dataflow.util.PackageUtil.stageOnePackage(PackageUtil.java:247)
	at org.apache.beam.runners.dataflow.util.PackageUtil.access$100(PackageUtil.java:65)
	at org.apache.beam.runners.dataflow.util.PackageUtil$2.run(PackageUtil.java:312)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at org.apache.beam.runners.dataflow.repackaged.com.google.common.util.concurrent.TrustedListenableFutureTask$TrustedFutureInterruptibleTask.runInterruptibly(TrustedListenableFutureTask.java:111)
	at org.apache.beam.runners.dataflow.repackaged.com.google.common.util.concurrent.InterruptibleTask.run(InterruptibleTask.java:58)
	at org.apache.beam.runners.dataflow.repackaged.com.google.common.util.concurrent.TrustedListenableFutureTask.run(TrustedListenableFutureTask.java:75)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
[...]

PR #1095, opened 2018-02-12
Default to "batch" if setBatchPath is never called (null).

jawein:
Fixes [#1073](https://github.com/google/google-api-java-client/issues/1073)

[#1087](https://github.com/google/google-api-java-client/pull/1087) aimed to fix this (and might have for some cases) but when I use it `batchPath` is `null` when not set and not an empty string.


jawein:
I signed it!

PR #1096, opened 2018-02-12
Handle case when batchPath is null

jawein:
Fixes [#1073](https://github.com/google/google-api-java-client/issues/1073)

[#1087](https://github.com/google/google-api-java-client/pull/1087) aimed to fix this (and might have for some cases) but when I use it `batchPath` is `null` when not set and not an empty string.


neildlg:
Will this fix be included the next release? I am stuck on version 2.1.0 because of the setBatchPath bug.

jadekler:
@neildlg Yes. We are working on releasing this project this week. (it takes quite some time unfortunately)

neildlg:
Hi, is there any update on when this fix will be released? From what I can tell there was a release in May but this fix wasn't included.

jadekler:
We are still persistently working on it, but I can't give a date. Hopefully soon!

PR #1106, opened 2018-04-12
Sync from google3

jadekler:


google-api-client/src/main/java/com/google/api/client/googleapis/media/MediaHttpDownloader.java
```diff
@@ -304,6 +303,22 @@ public MediaHttpDownloader setBytesDownloaded(long bytesDownloaded) {
    * @param lastBytePos The last byte position in the content range string.
    * @since 1.13
    */
+  public MediaHttpDownloader setContentRange(long firstBytePos, int lastBytePos) {
```
jadekler: Note: this was the result of a merge conflict (one function had long long, the other had long int). We decided to overload the method, but that might not have been the the correct choice.

jadekler:
Re-generating

PR #1111, opened 2018-05-03
internal: readme cleanup

jadekler:
Also document steps to build and test, incl getting google-play-services jar installed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ Platform Services.
 
 ```
 mkdir /tmp/foo && cd /tmp/foo
-wget --no-check-certificate https://dl.google.com/dl/android/maven2/com/google/android/gms/play-services-basement/8.3.0/play-services-basement-8.3.0.aar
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -10,15 +10,52 @@
...
 
-## <a name='Overview'>Overview<a/>
+## Building locally
 
-The Google APIs Client Library for Java is a flexible, efficient, and powerful Java client library for accessing any HTTP-based API on the web, not just Google APIs. 
+##### One time setup
+
+```
+mkdir /tmp/foo && cd /tmp/foo
+wget --no-check-certificate https://dl.google.com/dl/android/maven2/com/google/android/gms/play-services-basement/8.3.0/play-services-basement-8.3.0.aar
```
ejona86: Why `--no-check-certificate`?

jadekler: *shrug* Copy pasted what I had locally. I can remove, it's probably more canonical to let users figure that out themselves?

ejona86: Why do you need that locally? That means the TLS certs are broken, and isn't something that should generally be ignored.

jadekler: Removed the line + rebasing into master while we continue the conv offline

PR #1113, opened 2018-05-25
#1007 - adding option to impersonate a user with backward compatibility

netp-dany:
Overloading function createScoped in order to allow impersonation of a user.


netp-dany:
I signed it!

netp-dany:
For reference: https://github.com/google/google-api-java-client/issues/1007#

jadekler:
Thanks for the contribution.

This client library is supported but in maintenance mode only, unfortunately. We are mostly in a bug-fixing, security-patching mode. Furthermore, this library is currently frozen while we clean up some of the internals and prepare for a release. I'll keep this PR open and we'll circle back to this if we get some free cycles.

PR #1127, opened 2018-07-12
generated commit: switch to 1.25.0-SNAPSHOT

jadekler:


jadekler:
Note: CI is failing because it fails to find http-1.25.0 (because it's not released yet).

PR #1146, opened 2018-08-08
Add X-Goog-Api-Client header

chingor13:
To get better visibility into which Java versions/environments users are using, we want to add a header to all API requests that identify the client's language version (and other relevant environment versions).

See https://cloud.google.com/apis/docs/system-parameters

### DIFF
--- a/google-api-client/src/main/java/com/google/api/client/googleapis/services/AbstractGoogleClientRequest.java
+++ b/google-api-client/src/main/java/com/google/api/client/googleapis/services/AbstractGoogleClientRequest.java
@@ -12,6 +12,7 @@
 
 package com.google.api.client.googleapis.services;
 
+import com.google.api.client.googleapis.GoogleUtils;
 import com.google.api.client.googleapis.MethodOverride;
 import com.google.api.client.googleapis.batch.BatchCallback;
 import com.google.api.client.googleapis.batch.BatchRequest;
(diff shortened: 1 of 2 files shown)


google-api-client/src/main/java/com/google/api/client/googleapis/services/AbstractGoogleClientRequest.java
```diff
@@ -119,6 +124,48 @@ protected AbstractGoogleClientRequest(AbstractGoogleClient abstractGoogleClient,
...
+   * See <a href="https://cloud.google.com/apis/docs/system-parameters"></a>
+   *
+   */
+  private static class ApiClientVersion {
+    private static final String JAVA_VERSION = formatSemver(System.getProperty("java.version"));
+    private static final String OS_NAME = formatName(System.getProperty("os.name"));
+    private static final String OS_VERSION = formatSemver(System.getProperty("os.version"));
+
+    private static String build(AbstractGoogleClient client) {
+      // TODO(chingor): add the API version from the generated client
```
jadekler: I _think_ this is already happening.

jadekler: I _think_ this is somehow already happening, or partially happening. Happy to discuss offline if you care to investigate.

PR #1148, opened 2018-08-20
Generated commit - bump version to 1.26.0-SNAPSHOT

chingor13:


chingor13:
Tests are expected to fail because there is no published 1.26.0-SNAPSHOT version for http/oauth clients.

PR #1149, opened 2018-08-23
Add kokoro test configs

chingor13:
Also fixes the Java version reported in the header for Java 9. (which was inadvertently introduced in #1146)


chingor13:
This library does not currently have system tests.

theacodes:
Ah gotcha. Carry on!

PR #1166, opened 2018-09-18
Remove ClientLogin

### LINKED ISSUE #1161: Remove ClientLogin
specifically src/main/java/com/google/api/client/googleapis/auth/clientlogin/

This hasn't worked on the server for 3 years, and it all seems to be marked `@Beta.`

elharo:
@chingor13 fix #1161 This hasn't worked for 3 years now.


chingor13:
We still need this internally, but it looks like we can handle this legacy code internally. This PR isn't valid anymore as it targets dev instead of master and the upstream is no longer available.

I'll open a new PR that accomplishes this.

PR #1174, opened 2018-09-20
Fix Copyright notices

### LINKED ISSUE #1168: Bad Copyright Notices
Some files have a copyright notice like this:

Copyright (c) 2013 Google Inc.

This is incorrect under the law. (c) is meaningless. The word Copyright is sufficient. New files should have 

Copyright 2018 Google LLC.

though we don't have to change Inc. to LLC where it's otherwise fine.

elharo:
@chingor13 fix #1168


google-api-client-android/src/main/java/com/google/api/client/googleapis/extensions/android/accounts/GoogleAccountManager.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2012 Google Inc.
+ * Copyright 2012 Google Inc.
```
chingor13: Should they be `Google LLC` too?

elharo: I asked the internal copyright folks about this a few months back in another project. They told me not to bother changing that, but new files should be LLC.

PR #1183, opened 2018-09-21
we aren't using datanucleus anyway

### LINKED ISSUE #1167: Remove or replace DataNucleus
DataNucleus doesn't work very well on App Engine these days and the official recommendation is to replace it: https://cloud.google.com/appengine/docs/standard/java/datastore/jdo/overview-dn2

I'm not sure how deeply DataNucleus is woven into this project, but if it's not too tight we should get rid of it.

elharo:
@chingor13 Fix #1167 

Not sure how to regenerate google-api-client-assembly/dependencies/google-http-client-jdo-dependencies.html (and why are we checking generated code into the repo?)


elharo:
Please merge this @sduskis 

I do not have write access to this repo.

chingor13:
I was holding off because I don't know the implications of removing datanucleus. It rewrites to bytecode to provide the JDO interface, so I'm not sure if we can remove it (or perhaps we no longer want to provide the jdo artifact?).

PR #1184, opened 2018-09-21
Fix AppEngine tests

chingor13:
Updating the appengine sdk in #1178 introduced a new abstract method that was not implemented.


chingor13:
@elharo PTAL

google-api-client-appengine/src/main/java/com/google/api/client/googleapis/extensions/appengine/testing/auth/oauth2/MockAppIdentityService.java
```diff
@@ -89,4 +89,7 @@ public String getServiceAccountName() {
   public ParsedAppId parseFullAppId(String fullAppId) {
     return null;
   }
+
+  @Override
+  public String getDefaultGcsBucketName() { return null; }
```
elharo: three lines please

PR #1211, opened 2018-10-31
Delay request initialization for resumable upload

chingor13:
Wait until the content chunk is ready first.

Fixes internal bug b/118449379


chingor13:
cc @huygaa11

PR #1213, opened 2018-11-06
Add google-api-client-bom artifact

### LINKED ISSUE #1212: Create a BOM for google-api-java-client
Create a BOM for this library, same as the request for google-http-java-client at https://github.com/googleapis/google-http-java-client/issues/516 .

cc @elharo

chingor13:
Fixes #1212

### DIFF
--- a/google-api-client-bom/README.md
+++ b/google-api-client-bom/README.md
@@ -1,7 +1,7 @@
 # Google API Client Library Bill of Materials
 
-The `google-api-client-bom` modules is a pom that can be used to import consistent 
-versions of `google-api-client` components plus its dependencies.
+The `google-api-client-bom` module is a pom that can be used to import consistent 
+versions of `google-api-client` components.
 
 To use it in Maven, add the following to your `pom.xml`:
 
--- a/google-api-client-bom/README.md
+++ b/google-api-client-bom/README.md
@@ -1,0 +1,28 @@
+# Google API Client Library Bill of Materials
+
+The `google-api-client-bom` modules is a pom that can be used to import consistent 
+versions of `google-api-client` components plus its dependencies.
+
+To use it in Maven, add the following to your `pom.xml`:
+
+[//]: # ({x-version-update-start:google-api-client-bom:released})
+```xml
+<dependencyManagement>
+  <dependencies>
+    <dependency>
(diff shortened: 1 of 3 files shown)


google-api-client-bom/README.md
```diff
@@ -0,0 +1,28 @@
+# Google API Client Library Bill of Materials
+
+The `google-api-client-bom` modules is a pom that can be used to import consistent
```
garrettjonesgoogle:
```suggestion
The `google-api-client-bom` module is a pom that can be used to import consistent 
```

google-api-client-bom/README.md
```diff
@@ -0,0 +1,28 @@
+# Google API Client Library Bill of Materials
+
+The `google-api-client-bom` modules is a pom that can be used to import consistent 
+versions of `google-api-client` components plus its dependencies.
```
garrettjonesgoogle: It doesn't have the purpose of making sure consistent versions of dependencies are imported.

google-api-client-bom/pom.xml
```diff
@@ -0,0 +1,103 @@
...
+    <dependencies>
+      <dependency>
+        <groupId>com.google.api-client</groupId>
+        <artifactId>google-api-client</artifactId>
+        <version>1.26.1-SNAPSHOT</version><!-- {x-version-update:google-api-client:current} -->
+      </dependency>
+      <dependency>
+        <groupId>com.google.api-client</groupId>
+        <artifactId>google-api-client-android</artifactId>
+        <version>1.26.1-SNAPSHOT</version><!-- {x-version-update:google-api-client:current} -->
```
garrettjonesgoogle: (I only realized this after looking at the corresponding review for oauth) the artifactId is wrong in the x-version-update strings.

chingor13: In this repo, we're only tracking google-api-client, google-http-client, and google-oauth-client. We use the same x-version-update for simplicity as each artifact because they are all versioned the same.

google-api-client-bom/README.md
chingor13: Removed "plus its dependences"

google-api-client-bom/README.md
chingor13: Done

PR #1221, opened 2018-12-11
Remove usage of deprecated Backoff from google-http-java-client

chingor13:
BackoffPolicy was deprecated in google-http-java-client which was scheduled for deletion in March 2014!


sduskis:
Does this mean that by default users will not use exponential backoff?

chingor13:
There is no default backoff set for users. If they are explicitly setting the deprecated `backOffPolicy()` in their `RequestFactory` or `HttpRequestInitializer`, it will now be ignored.

PR #1238, opened 2019-01-29
3929: Changed to Guava ByteStreams.copy() that has a bit better performance

andrey-qlogic:
Fixes #3929

- [ ] Tests pass
- [ ] Appropriate docs were updated (if necessary)


andrey-qlogic:
wrong commiter email

PR #1239, opened 2019-01-29
3929: Changed to Guava ByteStreams.copy() that has a bit better performance

andrey-qlogic:
Fixes https://github.com/googleapis/google-cloud-java/issues/3929


andrey-qlogic:
@frankyn , would you please take a look.

andrey-qlogic:
IOUtils.copy() uses BUF_SIZE = 4096, but Guava ByteStreams.copy 8192. That allows to copy slightly faster.

frankyn:
Thanks Andrey, that makes sense. Given it's the default value, did you try using the overload for [copy(java.io.InputStream, java.io.OutputStream, int)](https://commons.apache.org/proper/commons-io/javadocs/api-2.5/org/apache/commons/io/IOUtils.html#copy(java.io.InputStream,%20java.io.OutputStream,%20int)) that adjusts the size of the copy? 

If the issue is the buffer size then the code could use the same library and use the overload instead.

PR #1243, opened 2019-02-21
Fix Replaced invalidateToken method to clearToken

### LINKED ISSUE #950: Replace GoogleAuthUtil.invalidateToken with GoogleAuthUtil.clearToken
It's used here:

https://github.com/google/google-api-java-client/blob/57fe35766cbba0a0d5a9a296be81468d730a29f8/google-api-client-android/src/main/java/com/google/api/client/googleapis/extensions/android/gms/auth/GoogleAccountCredential.java#L294

The method is deprecated due to requiring MANAGE_ACCOUNTS and USE_CREDENTIALS permissions:

https://developers.google.com/android/reference/com/google/android/gms/auth/GoogleAuthUtil#invalidateToken(android.content.Context, java.lang.String)

The replacement is GoogleAuthUtil.clearToken. This should allow Android developers to use the library without requesting those extra permissions.

pmakani:
Fixes #950

### DIFF
--- a/google-api-client-android/src/main/java/com/google/api/client/googleapis/extensions/android/gms/auth/GoogleAccountCredential.java
+++ b/google-api-client-android/src/main/java/com/google/api/client/googleapis/extensions/android/gms/auth/GoogleAccountCredential.java
@@ -301,8 +301,8 @@ public class GoogleAccountCredential implements HttpRequestInitializer {
     }
 
     @Override
-    public boolean handleResponse(
-        HttpRequest request, HttpResponse response, boolean supportsRetry) {
+    public boolean handleResponse(HttpRequest request, HttpResponse response, boolean supportsRetry)
+        throws IOException {
       try {
         if (response.getStatusCode() == 401 && !received401) {
           received401 = true;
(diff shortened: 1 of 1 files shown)


ladyking113:
Yes

google-api-client-android/src/main/java/com/google/api/client/googleapis/extensions/android/gms/auth/GoogleAccountCredential.java
```diff
@@ -303,10 +303,16 @@ public void intercept(HttpRequest request) throws IOException {
...
-        GoogleAuthUtil.invalidateToken(context, token);
-        return true;
+      try {
+        if (response.getStatusCode() == 401 && !received401) {
+          received401 = true;
+          GoogleAuthUtil.clearToken(context, token);
+          return true;
+        }
+      } catch (GoogleAuthException e) {
+        e.printStackTrace();
```
sduskis: I don't think we can just swallow the exception here.  Why would these exceptions be thrown?

pmakani: Resolved.

PR #1251, opened 2019-05-10
merge OSGI metadata

elharo:
This fixes OSGI metadata for the jackson2 package. Others still need to be done. 

- [X ] Tests pass
- [ X] Appropriate docs were updated (if necessary)

### DIFF
--- a/google-api-client-gson/pom.xml
+++ b/google-api-client-gson/pom.xml
@@ -26,13 +26,35 @@
         <artifactId>maven-jar-plugin</artifactId>
         <configuration>
           <archive>
+            <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
             <manifest>
               <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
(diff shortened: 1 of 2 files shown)


google-api-client-gson/pom.xml
```diff
@@ -26,15 +26,37 @@
         <artifactId>maven-jar-plugin</artifactId>
         <configuration>
           <archive>
+            <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
             <manifest>
               <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
             </manifest>
             <manifestEntries>
-              <Automatic-Module-Name>google.api.client.gson</Automatic-Module-Name>
+              <Automatic-Module-Name>google.api.client.jackson2</Automatic-Module-Name>
```
chingor13: This doesn't seem right. It should probably stay `gson`

elharo: yep, done

PR #1259, opened 2019-05-16
Update http/oauth dependencies to 1.29

chingor13:
* Bump oauth and http client dependencies.
* Stop version bumping for oauth and http clients at release time


chingor13:
@sduskis Fixed the build, PTAL

PR #1263, opened 2019-05-22
Extensible client options

chingor13:
Introduces the builder pattern for these common configuration options so that it is extensible in the future.
Add ability to configure `requestReason` and `userAgent`.


sduskis:
Should this class use Auto-Value?

chingor13:
> Should this class use Auto-Value?

We certainly could. We're not using autovalue yet for any of the other builders (or at all). It might make sense to come back with a refactoring PR to convert the all builders to use autovalue (if it's even possible). Otherwise, we will have a mix of several manually defined builders and one autovalue builder. WDYT?

chingor13:
@sduskis Stubbed out an autovalue implementation of `CommonGoogleOptions`. I'm not sure we can shim autovalue into the existing classes as they are already part of a class where users instantiate the classes directly.

chingor13:
I'm not sold on trying to shoehorn autovalue in here. If we were starting from scratch, it would make a lot of sense. We cannot switch many of these implementations over to autovalue as they already have public constructors (including the `CommonGoogleClientRequestInitializer`)

PR #1323, opened 2019-06-13
Add option to return raw input stream for response

ajaaym:
Add option to return raw input stream for response

### DIFF
--- a/google-api-client/src/main/java/com/google/api/client/googleapis/services/AbstractGoogleClientRequest.java
+++ b/google-api-client/src/main/java/com/google/api/client/googleapis/services/AbstractGoogleClientRequest.java
@@ -94,6 +94,9 @@ public abstract class AbstractGoogleClientRequest<T> extends GenericData {
   /** Whether to disable GZip compression of HTTP content. */
   private boolean disableGZipContent;
 
+  /** whether to return raw input stream in {@link HttpResponse#getContent()}. */
+  private boolean returnRawInputStream;
+
   /** Response class to parse into. */
   private Class<T> responseClass;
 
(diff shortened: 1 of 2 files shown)


ajaaym:
@chingor13 can you please review this, this is towards [3929](https://github.com/googleapis/google-cloud-java/issues/3929)

google-api-client/src/main/java/com/google/api/client/googleapis/services/AbstractGoogleClientRequest.java
```diff
@@ -239,6 +247,25 @@ public final boolean getDisableGZipContent() {
     return this;
   }
 
+  /**
+   * Sets whether the response should return raw input stream or not.
+   *
+   * <p>
+   * By default ti is {@code false}.
```
chingor13:
```suggestion
   * By default it is {@code false}.
```

google-api-client/src/main/java/com/google/api/client/googleapis/services/AbstractGoogleClientRequest.java
```diff
@@ -94,6 +94,9 @@
   /** Whether to disable GZip compression of HTTP content. */
   private boolean disableGZipContent;
 
+  /** whether to return raw input stream in {@link HttpResponse#getContent()}. */
```
chingor13:
```suggestion
  /** Whether to return raw input stream in {@link HttpResponse#getContent()}. */
```

google-api-client/src/main/java/com/google/api/client/googleapis/services/AbstractGoogleClientRequest.java
```diff
@@ -239,6 +247,25 @@ public final boolean getDisableGZipContent() {
...
 
+  /**
+   * Sets whether the response should return raw input stream or not.
+   *
+   * <p>
+   * By default ti is {@code false}.
+   * </p>
+   *
+   * <p>
+   * Normally response stream is wrapped GZipInputStream when it contains content-encoding header
```
chingor13: Maybe:

When the response contains a known content-encoding header, the response stream is wrapped with an InputStream that decodes the content. This fails when we download large files in chunks (see #bugnumber). Setting this to true will make the response return the raw input stream.

ajaaym:
@chingor13 not sure whats going on with cla/google, can you check?

PR #1324, opened 2019-06-17
Close streams on error

SUPERCILEX:


SUPERCILEX:
cc @ajaaym

ajaaym:
@SUPERCILEX can you please describe the issue for this? And if this cant be handled in application code? This library is used by googleapis/google-cloud-java and we need to see how it affects there in terms of retries before we can merge this.

SUPERCILEX:
I don't understand how this could possibly be the client's responsibility. Here's the scenario:
1. Start uploading a file. For example, with Gradle Play Publisher: https://github.com/Triple-T/gradle-play-publisher/blob/2181b5bafd32d5f7554dcbd5bfd07505e2b39be2/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/PublishListing.kt#L228-L234
1. The upload fails with some error code (500 for example).
1. Now, the file will still be open and the only way to close it is to kill the process who opened it.

AFAIK, you can't go backwards in a stream so the stream will broken anyway if someone tries to continue using it.

SUPERCILEX:
@ajaaym @sduskis is there anything I can do to speed the process up?

ajaaym:
@SUPERCILEX it seems ok to merge, @chingor13 can you please review?

chingor13:
Yeah, I was looking into writing a test for it. The test setup for the media upload is pretty involved.

chingor13:
Thank you!

SUPERCILEX:
Awesome, thank you! 👏

PR #1333, opened 2019-06-27
Deprecate BatchRequest constructor

chingor13:
Users should obtain a `BatchRequest` instance via their client's `batch(HttpRequestInitializer)` method.

* Deprecates the public BatchRequest constructor
* Logs a WARNING message if you try to make a request against the global batch endpoint.

### DIFF
--- a/google-api-client/src/main/java/com/google/api/client/googleapis/batch/BatchRequest.java
+++ b/google-api-client/src/main/java/com/google/api/client/googleapis/batch/BatchRequest.java
@@ -32,6 +32,8 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 /**
  * An instance of this class represents a single batch of requests.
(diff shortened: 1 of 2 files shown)


google-api-client/src/main/java/com/google/api/client/googleapis/batch/BatchRequest.java
```diff
@@ -213,6 +230,13 @@ public int size() {
   public void execute() throws IOException {
     boolean retryAllowed;
     Preconditions.checkState(!requestInfos.isEmpty());
+
+    // Log a warning if the user is using the global batch endpoint. In the future, we can turn this
+    // into a preconditions check.
+    if (this.batchUrl.toString().equals(GLOBAL_BATCH_ENDPOINT)) {
```
kolea2: I would flip the equality here to avoid a NPE.

google-api-client/src/main/java/com/google/api/client/googleapis/services/AbstractGoogleClient.java
```diff
@@ -232,6 +232,7 @@ public final BatchRequest batch() {
    * @return newly created Batch request
    */
   public final BatchRequest batch(HttpRequestInitializer httpRequestInitializer) {
+    @SuppressWarnings("deprecated")
     BatchRequest batch =
```
kolea2: should we migrate this to the new API?

kolea2: Should we migrate this to the new API?

chingor13: This one is the method that users should be using that sets the batch path.

google-api-client/src/main/java/com/google/api/client/googleapis/batch/BatchRequest.java
```diff
@@ -94,8 +97,19 @@ public void onFailure(GoogleJsonErrorContainer e, HttpHeaders responseHeaders) {
  */
 public final class BatchRequest {
 
+  /**
+   * The deprecated global batch endpoint. Users should actually use the per-service batch endpoint
+   * declared by the service configuration.
+   */
+  private static final String GLOBAL_BATCH_ENDPOINT = "https://www.googleapis.com/batch";
+  private static final String GLOBAL_BATCH_ENDPOINT_WARNING = "You are using the global batch "
```
kolea2: Maybe we can also link to javadoc or a more verbose example?

google-api-client/src/main/java/com/google/api/client/googleapis/batch/BatchRequest.java
chingor13: Done.

google-api-client/src/main/java/com/google/api/client/googleapis/batch/BatchRequest.java
chingor13: We would like to migrate the docs off of developers.google.com for these clients, so I migrated the batching and media upload/download docs to the main README in #1334. That will give us a place to target with this message.

chingor13: Added a link to the new docs section in the warning message.

PR #1355, opened 2019-08-12
Fix VERSION constant

chingor13:
Add markers for version replacement so this constant is kept up to date via releasetool.

See https://github.com/googleapis/releasetool/pull/216

### DIFF
--- a/google-api-client/src/main/java/com/google/api/client/googleapis/GoogleUtils.java
+++ b/google-api-client/src/main/java/com/google/api/client/googleapis/GoogleUtils.java
@@ -15,11 +15,12 @@
 package com.google.api.client.googleapis;
 
 import com.google.api.client.util.SecurityUtils;
-
 import java.io.IOException;
 import java.io.InputStream;
 import java.security.GeneralSecurityException;
 import java.security.KeyStore;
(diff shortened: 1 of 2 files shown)


google-api-client/src/main/java/com/google/api/client/googleapis/GoogleUtils.java
```diff
@@ -29,42 +30,50 @@
...
    */
-  public static final Integer BUGFIX_VERSION = 0;
+  public static final Integer BUGFIX_VERSION;
 
-  /** Current release version. */
-  // NOTE: toString() so compiler thinks it isn't a constant, so it won't inline it
-  public static final String VERSION = (MAJOR_VERSION + "." + MINOR_VERSION + "." + BUGFIX_VERSION
-      + "-SNAPSHOT").toString();
+  static {
+    Matcher versionMatcher =
```
kolea2: Would just add a small test for the matcher.

chingor13: Done. Moved the pattern into a package private constant visible for testing.

PR #1395, opened 2019-10-20
Add details to GoogleJsonResponseExceptions created with GoogleJsonResponseExceptionFactoryTesting

SUPERCILEX:
Currently, `GoogleJsonResponseException`s created with `GoogleJsonResponseExceptionFactoryTesting` will have null `#getDetails()`.

### DIFF
--- a/google-api-client/src/main/java/com/google/api/client/googleapis/testing/json/GoogleJsonResponseExceptionFactoryTesting.java
+++ b/google-api-client/src/main/java/com/google/api/client/googleapis/testing/json/GoogleJsonResponseExceptionFactoryTesting.java
@@ -17,6 +17,7 @@ package com.google.api.client.googleapis.testing.json;
 import com.google.api.client.googleapis.json.GoogleJsonResponseException;
 import com.google.api.client.http.HttpRequest;
 import com.google.api.client.http.HttpResponse;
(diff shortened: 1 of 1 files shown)


google-api-client/src/main/java/com/google/api/client/googleapis/testing/json/GoogleJsonResponseExceptionFactoryTesting.java
```diff
@@ -59,7 +59,10 @@ public static GoogleJsonResponseException newMock(JsonFactory jsonFactory,
     MockLowLevelHttpResponse otherServiceUnavaiableLowLevelResponse =
         new MockLowLevelHttpResponse()
         .setStatusCode(httpCode)
-        .setReasonPhrase(reasonPhrase);
+        .setReasonPhrase(reasonPhrase)
+        .setContentType(Json.MEDIA_TYPE)
```
chingor13: Looks like we're missing an import here or a constant definition.

SUPERCILEX: Whoops, fixed.

SUPERCILEX:
Awesome, thanks!

PR #1418, opened 2019-11-12
chore(docs): remove out of date docs

elharo:
@chingor13  Please double check me, but I don't think we need any of this and it's all out of date and mostly autogenerated.


chingor13:
The dependencies files are included into the assembly zip. It's possible that we can use maven-site-plugin to try and generate the dependencies report and include those in the assembly.

google-api-client-assembly/dependencies/APACHE-LICENSE.txt
```diff
@@ -1,202 +0,0 @@
-
```
chingor13, line 1: We probably still need to keep the license files in here. These are the collected license files of the dependencies used and are required if we redistribute.

chingor13 (changes requested):

PR #1425, opened 2019-11-20
fix: user-agent should use identifier/version

### LINKED ISSUE #1426: User-Agent string uses invalid format
The default user agent (and appended user agent) only uses `Google-API-Java-Client`. It should include the version (e.g. `Google-API-Java-Client/1.30.6`)

This blocks reporting on the backend with regards to which version of the google-api-client is being used.

chingor13:
Fixes #1426

### DIFF
--- a/google-api-client/src/test/java/com/google/api/client/googleapis/services/AbstractGoogleClientRequestTest.java
+++ b/google-api-client/src/test/java/com/google/api/client/googleapis/services/AbstractGoogleClientRequestTest.java
@@ -204,7 +204,7 @@ public class AbstractGoogleClientRequestTest extends TestCase {
     // Specify an Application Name.
     String applicationName = "Test Application";
     transport.expectedUserAgent = applicationName + " "
-        + AbstractGoogleClientRequest.USER_AGENT_SUFFIX + "/" + GoogleUtils.VERSION + " "
+        + "Google-API-Java-Client/" + GoogleUtils.VERSION + " "
         + HttpRequest.USER_AGENT_SUFFIX;
     MockGoogleClient client = new MockGoogleClient.Builder(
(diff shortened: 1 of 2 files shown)


google-api-client/src/test/java/com/google/api/client/googleapis/services/AbstractGoogleClientRequestTest.java
```diff
@@ -203,7 +204,7 @@ public void testUserAgentSuffix() throws Exception {
     // Specify an Application Name.
     String applicationName = "Test Application";
     transport.expectedUserAgent = applicationName + " "
```
elharo: You can inline AbstractGoogleClientRequest.USER_AGENT_SUFFIX  and applicationName here and the test becomes a little stronger since it's no longer duplicating as much of the logic in the model code.

chingor13: That constant is public, so I'm not sure if that would possibly break users.

chingor13: Oh, misread the comment. I'll update that.

elharo: Using a string literal instead of a constant tests the constant. Using the constant simply guarantees that any mistakes in the constant value will not be caught by the tests. Yes, I've seen constants with bugs in them. :-(

chingor13: Done.

PR #1452, opened 2019-12-17
fix: update GoogleUtils#getVersion() to use stable logic

### LINKED ISSUE #1451: Android/Java 1.30.6 causing crash due to PR #1419
Prompted by this SO post with more details: https://stackoverflow.com/questions/59243821/youtube-api-exception-in-initializer-error/59298544#59298544

#### Environment details

- OS: Android
- Java version: 8
- google-api-java-client version: 1.30.6

#### Steps to reproduce

1. Try to access GoogleUtils.VERSION with build.gradle looking like:
`dependencies {
  implementation ('com.google.api-client:google-api-client:1.30.6')
  implementation ('com.google.api-client:google-api-client-android:1.30.6')
  implementation ('com.google.apis:google-api-services-drive:+')
}` 

#### Stacktrace
```
Caused by: java.lang.IllegalStateException: No successful match so far
        at java.util.regex.Matcher.ensureMatch(Matcher.java:1116)
        at java.util.regex.Matcher.group(Matcher.java:382)
        at com.google.api.client.googleapis.GoogleUtils.<clinit>(Unknown Source:26)
```

Please let me know if this is a problem with our own setups, as far as I can tell there's not a whole lot I can do.

Thanks!

chingor13:
Fixes #1451

### DIFF
--- a/google-api-client/src/main/java/com/google/api/client/googleapis/GoogleUtils.java
+++ b/google-api-client/src/main/java/com/google/api/client/googleapis/GoogleUtils.java
@@ -91,18 +91,18 @@ public final class GoogleUtils {
   private static String getVersion() {
     // attempt to read the library's version from a properties file generated during the build
     // this value should be read and cached for later use
-    String version = "unknown-version";
+    String version = null;
     try (InputStream inputStream =
         GoogleUtils.class.getResourceAsStream("google-api-client.properties")) {
       if (inputStream != null) {
-        final Properties properties = new Properties();
+        Properties properties = new Properties();
         properties.load(inputStream);
         version = properties.getProperty("google-api-client.version");
       }
     } catch (IOException e) {
       // ignore
     }
-    return version;
+    return version == null ? "unknown-version" : version;
   }
 
   private GoogleUtils() {}
--- a/google-api-client/src/main/java/com/google/api/client/googleapis/GoogleUtils.java
+++ b/google-api-client/src/main/java/com/google/api/client/googleapis/GoogleUtils.java
@@ -89,20 +89,18 @@ public final class GoogleUtils {
   }
 
   private static String getVersion() {
-    String version = GoogleUtils.class.getPackage().getImplementationVersion();
-    // in a non-packaged environment (local), there's no implementation version to read
-    if (version == null) {
-      // fall back to reading from a properties file - note this value is expected to be cached
-      try (InputStream inputStream =
-          GoogleUtils.class.getResourceAsStream("google-api-client.properties")) {
-        if (inputStream != null) {
-          Properties properties = new Properties();
-          properties.load(inputStream);
-          version = properties.getProperty("google-api-client.version");
-        }
-      } catch (IOException e) {
-        // ignore
+    // attempt to read the library's version from a properties file generated during the build
+    // this value should be read and cached for later use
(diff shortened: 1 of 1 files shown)


google-api-client/src/main/java/com/google/api/client/googleapis/GoogleUtils.java
```diff
@@ -89,20 +89,18 @@ public static synchronized KeyStore getCertificateTrustStore()
...
+    String version = "unknown-version";
+    try (InputStream inputStream =
+        GoogleUtils.class.getResourceAsStream("google-api-client.properties")) {
+      if (inputStream != null) {
+        final Properties properties = new Properties();
+        properties.load(inputStream);
+        version = properties.getProperty("google-api-client.version");
       }
+    } catch (IOException e) {
+      // ignore
```
elharo: do we really want to ignore this and not fill in a default?

google-api-client/src/main/java/com/google/api/client/googleapis/GoogleUtils.java
```diff
@@ -89,20 +89,18 @@ public static synchronized KeyStore getCertificateTrustStore()
...
-        }
-      } catch (IOException e) {
-        // ignore
+    // attempt to read the library's version from a properties file generated during the build
+    // this value should be read and cached for later use
+    String version = "unknown-version";
+    try (InputStream inputStream =
+        GoogleUtils.class.getResourceAsStream("google-api-client.properties")) {
+      if (inputStream != null) {
+        final Properties properties = new Properties();
```
elharo: probably doesn't need to be final

google-api-client/src/main/java/com/google/api/client/googleapis/GoogleUtils.java
elharo: NVM, that's what we do effectively.

google-api-client/src/main/java/com/google/api/client/googleapis/GoogleUtils.java
```diff
@@ -89,20 +89,18 @@ public static synchronized KeyStore getCertificateTrustStore()
...
+        properties.load(inputStream);
+        version = properties.getProperty("google-api-client.version");
```
elharo: this could be null. You should supply a default here.

chingor13: Done. Moved the default to the return statement while checking for null.

elharo: Or you could write

```
version = properties.getProperty("google-api-client.version". "unknown-version");
```

up to you.

Samikay:
Did this make it into 1.30.7? Cause it's still broken.

     Caused by: java.lang.IllegalStateException: No successful match so far
        at java.util.regex.Matcher.ensureMatch(Matcher.java:1116)
        at java.util.regex.Matcher.group(Matcher.java:382)
        at c.c.b.a.c.a.<clinit>(:66)
        at com.google.api.services.drive.Drive.<clinit>(:44) 
        at com.google.api.services.drive.Drive$Builder.build(:9813)

elharo:
related #1467

PR #1457, opened 2019-12-18
fix: future test breakage

codyoss:
In a a future commit the wrapping of these InputStreams will change.
This should ensure the tests will still pass while at the same time
validating the raw InputStream behavior still works.


google-api-client/src/test/java/com/google/api/client/googleapis/services/AbstractGoogleClientRequestTest.java
```diff
@@ -308,7 +306,8 @@ public LowLevelHttpResponse execute() {
         client, HttpMethods.GET, URI_TEMPLATE, null, String.class);
     request.setReturnRawInputStream(true);
     InputStream inputStream = request.executeAsInputStream();
-    assertFalse(inputStream instanceof GZIPInputStream);
+    // The response will not be wrapped do to setReturnRawInputStream(true)
```
elharo: due to?

codyoss: Yup, thank you.

Issue #1487, opened 2020-02-03
Unresolvable dependency after 1.30.8


worpet:
#### Environment details
1. OS type and version: Linux
2. Java version: 8
3. google-api-client version(s): 1.30.8

#### Steps to reproduce
1. Attempt to build and resolve dependencies with Ivy.
2. Ivy fails to resolve `androidx.annotation#annotation;1.1.0`.

#### Code example

```
<dependency org="com.google.api-client" name="google-api-client" rev="1.30.8" conf="compile->default"/>
```

#### Stack trace
```
:: problems summary ::
:::: WARNINGS
		module not found: androidx.annotation#annotation;1.1.0

	==== central: tried

	  https://repo1.maven.org/maven2/androidx/annotation/annotation/1.1.0/annotation-1.1.0.pom

	  -- artifact androidx.annotation#annotation;1.1.0!annotation.jar:

	  https://repo1.maven.org/maven2/androidx/annotation/annotation/1.1.0/annotation-1.1.0.jar

		::::::::::::::::::::::::::::::::::::::::::::::

		::          UNRESOLVED DEPENDENCIES         ::

		::::::::::::::::::::::::::::::::::::::::::::::

		:: androidx.annotation#annotation;1.1.0: not found

		::::::::::::::::::::::::::::::::::::::::::::::
```

#### Any additional information below
Previous versions up to 1.30.7 build without any problems.

This seems to mention a similar error, but I am not positive if it is related: https://github.com/googleapis/google-api-java-client/pull/1467#issuecomment-572505559
> [ERROR] /tmpfs/src/github/google-api-java-client/google-api-client/src/main/java/com/google/api/client/googleapis/GoogleUtils.java:[17,26] error: package androidx.annotation does not exist

chingor13:
https://github.com/googleapis/google-api-java-client/pull/1467 looks like it introduced a dependency that exists in Google's Android Maven repository. I'm not sure that we should have taken this dependency as it seems to force all downstream users to add a Maven repository.

As a workaround, you can add `https://maven.google.com/` as an additional Maven repository in your project configuration.

elharo:
There was a very specific bug fixed by this change. We can explore other options but I don't think requiring that repository is as bad as the bug it fixes.

chingor13:
This seems to be only affecting ivy and not Maven or gradle users. Please try adding the additional Maven repository url.

worpet:
[...]

PR #1467, opened 2020-01-08
keep GoogleUtils

elharo:
Fixes #1450 

@chingor13

PR #1488, opened 2020-02-10
revert: keep GoogleUtils (#1467)

chingor13:
This reverts commit b70e8ef1145fad30bfe02e91d4dc9e278fd2d519.

#1487


elharo:
alternate solution found based on R8 so this is moot

PR #1491, opened 2020-02-14
fix: use embedded Proguard configuration instead of compile-time annotation

### LINKED ISSUE #1487: Unresolvable dependency after 1.30.8

anuraaga:
R8 provides a mechanism for embedding proguard files, which obviates the need for a compile-time dependency on an Android library in core.

Not well documented but reasonably commonly used.

https://twitter.com/jakewharton/status/1004401938467876865?lang=en
https://github.com/square/retrofit/blob/master/retrofit/src/main/resources/META-INF/proguard/retrofit2.pro

Unfortunately, without a regression test I'm not 100% sure this works, nor am I an Android developer with much confidence, just following the pattern of other libraries :)

Fixes #1487

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ To use Gradle, add the following lines to your build.gradle file:
 ```gradle
 repositories {
     mavenCentral()
+    google()
 }
 dependencies {
     compile 'com.google.api-client:google-api-client:1.30.8'
--- a/README.md
+++ b/README.md
@@ -45,7 +45,6 @@ To use Gradle, add the following lines to your build.gradle file:
 ```gradle
 repositories {
     mavenCentral()
-    google()
 }
 dependencies {
     compile 'com.google.api-client:google-api-client:1.30.8'
--- a/google-api-client/pom.xml
+++ b/google-api-client/pom.xml
@@ -145,23 +145,10 @@
       <artifactId>commons-codec</artifactId>
       <scope>provided</scope>
       <version>1.14</version>
-    </dependency>
+   </dependency>
     <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
     </dependency>
-    <dependency>
-      <groupId>androidx.annotation</groupId>
-      <artifactId>annotation</artifactId>
-      <version>1.1.0</version>
-    </dependency>
   </dependencies>
-  
-  <repositories>  
-    <repository>
-      <id>google</id>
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -45,7 +45,6 @@ To use Gradle, add the following lines to your build.gradle file:
 ```gradle
 repositories {
     mavenCentral()
-    google()
```
elharo: we probably want to leave this here until we're ready to release 1.30.9

anuraaga: Ah yeah - reverted!

anuraaga:
Not sure of versions - I don't see any reason for Android API level to be affected since this is purely compilation but there might be an SDK tooling implication.

@JakeWharton do you happen to know?

IanGClifton:
I can at least confirm that it won't impact Android OS versions, but I don't know if this works when not compiling with R8.

PR #1505, opened 2020-03-09
chore(deps): update project.http.version to v1.36.0

medb:


google-api-client/src/main/java/com/google/api/client/googleapis/json/GoogleJsonResponseException.java
```diff
@@ -141,6 +141,9 @@ public static GoogleJsonResponseException from(JsonFactory jsonFactory, HttpResp
     }
     // message
     StringBuilder message = HttpResponseException.computeMessageBuffer(response);
+    (message.length() == 0 ? message : message.append(StringUtils.LINE_SEPARATOR))
+        .append("Request URL: ")
```
medb: Would it better to just append an URL on a new line without `Request URL: ` prefix?

google-api-client/src/main/java/com/google/api/client/googleapis/json/GoogleJsonResponseException.java
```diff
@@ -141,6 +141,9 @@ public static GoogleJsonResponseException from(JsonFactory jsonFactory, HttpResp
     }
     // message
     StringBuilder message = HttpResponseException.computeMessageBuffer(response);
+    (message.length() == 0 ? message : message.append(StringUtils.LINE_SEPARATOR))
```
elharo: It took me some effort to decode this. It would be clearer as a simple if clause with no ternary operator.

medb: Done.

google-api-client/src/test/java/com/google/api/client/googleapis/json/GoogleJsonResponseExceptionTest.java
```diff
@@ -55,7 +56,13 @@ public void testFrom_withDetails() throws Exception {
         GoogleJsonResponseException.from(GoogleJsonErrorTest.FACTORY, response);
     assertEquals(GoogleJsonErrorTest.ERROR, GoogleJsonErrorTest.FACTORY.toString(ge.getDetails()));
     assertTrue(
-        ge.getMessage(), ge.getMessage().startsWith("403" + StringUtils.LINE_SEPARATOR + "{"));
+        ge.getMessage(),
```
elharo: ge should probably be unabbreviated. Reading it here, I have no idea what this variable is.

google-api-client/src/main/java/com/google/api/client/googleapis/json/GoogleJsonResponseException.java
medb: @elharo what do you think about this? Should we include `Request URL:` prefix in the error message?

```
500 Server error
http://storage.googleapis.com/bucket/...
```

vs

```
500 Server error
Request URL: http://storage.googleapis.com/bucket/...
```

elharo: I don't have a strong opinion here. Maybe include it? Either way is fine with me.

medb: Ack, will leave it as is.

google-api-client/src/test/java/com/google/api/client/googleapis/json/GoogleJsonResponseExceptionTest.java
medb: Done.

google-api-client/src/main/java/com/google/api/client/googleapis/json/GoogleJsonResponseException.java
```diff
@@ -141,6 +141,10 @@ public static GoogleJsonResponseException from(JsonFactory jsonFactory, HttpResp
     }
     // message
     StringBuilder message = HttpResponseException.computeMessageBuffer(response);
+    if (message.length() > 0) {
+      message.append(StringUtils.LINE_SEPARATOR);
```
elharo: The message should be platform independent. Just use `\n`.

medb: Done.

google-api-client/src/main/java/com/google/api/client/googleapis/json/GoogleJsonResponseException.java
```diff
@@ -141,6 +141,10 @@ public static GoogleJsonResponseException from(JsonFactory jsonFactory, HttpResp
     }
     // message
     StringBuilder message = HttpResponseException.computeMessageBuffer(response);
```
medb: Maybe I should make this change in the [`HttpResponseException. computeMessageBuffer`](https://github.com/googleapis/google-http-java-client/blob/027c227e558164f77be204152fb47023850b543f/google-http-client/src/main/java/com/google/api/client/http/HttpResponseException.java#L274) method in [google-http-java-client](https://github.com/googleapis/google-http-java-client) instead? @elharo wdyt?

elharo: That could work.

medb: Send PR: https://github.com/googleapis/google-http-java-client/pull/1002

medb: Re-purposed this PR for upgrading to newer HTTP client (when it will be released) that has request method in URL in the exception message.

PR #1536, opened 2020-06-17
replace deprecated apache transport with recommended

ckiosidis:
Fixes #<issue_number_goes_here> (it's a good idea to open an issue first for context and/or discussion)

Replaced the ApacheHttpTransport based on the recommendation at: https://github.com/googleapis/google-http-java-client/blob/master/google-http-client/src/main/java/com/google/api/client/http/apache/ApacheHttpTransport.java#L77-L78


chingor13:
If we want to do something similar, we should create a v2 version and deprecate this existing one. See #1568

PR #1619, opened 2020-11-01
feat: add mtls support to GoogleNetHttpTransport and GoogleApacheHttpTransport

arithmetic1728:


google-api-client/pom.xml
```diff
@@ -154,5 +159,17 @@
       <groupId>com.google.http-client</groupId>
       <artifactId>google-http-client-apache-v2</artifactId>
     </dependency>
+    <dependency>
```
elharo, line 162: We use Mockito for all new tests

elharo (changes requested):

google-api-client/pom.xml
arithmetic1728: I need to mock static method and environment variables, it seems Mockito cannot do either of these.

elharo: Mockito can mock static methods as of version 3.4.

arithmetic1728: To mock static method with Mockito, it seems we need to either pass in a function reference or a lambda to `when`, but none of these are supported in Java 7.

```
try(MockedStatic<Utils> utilsMock = Mockito.mockStatic(Utils.class)) {
      // use method reference inside "when"
      utilsMock.when(Utils::loadDefaultCertificate).thenReturn(mtlsKeyStore);
}
```
```
try (MockedStatic<System> systemMock = Mockito.mockStatic(System.class)) {
      // use lambda inside "when"
      systemMock.when(() -> System.getenv("foo")).thenReturn("true");
}
```

google-api-client/src/main/java/com/google/api/client/googleapis/apache/v2/GoogleApacheHttpTransport.java
```diff
@@ -41,9 +43,31 @@
...
+   * If `GOOGLE_API_USE_CLIENT_CERTIFICATE` environment variable is set to "true",
+   * the function looks for user provided client certificate first from 
+   * clientCertificateSource InputStream, if not exists, then the default from
+   * {@link Utils#loadDefaultCertificate()}. If client certificate exists,
+   * the transport uses it and is mutual TLS.
+   * 
+   * @param clientCertificateSource InputStream for mutual TLS client certificate and private key   
+   */
+  @Beta
+  public static ApacheHttpTransport newTrustedTransport(InputStream clientCertificateSource)
```
chingor13, line 69: I don't think we need to make this public. If we need to allow the user to customize the mtls credentials, we should directly pass through the `KeyStore` and password without requiring an `InputStream`. We are already providing helpers to load a `KeyStore` from `InputStream`.

You can use the default visibility if you want to dependency inject for testing.

google-api-client/pom.xml
```diff
@@ -117,6 +117,11 @@
     </resources>
   </build>
   <dependencies>
+    <dependency>
+      <groupId>com.google.code.gson</groupId>
+      <artifactId>gson</artifactId>
```
chingor13, line 122: This library already depends on jackson2 and we don't need 2 JSON parsing/serialization libraries at the same time. We should pick one and stick to it.

google-api-client/src/main/java/com/google/api/client/googleapis/javanet/GoogleNetHttpTransport.java
```diff
@@ -51,10 +58,54 @@ static HttpTransport newProxyTransport() throws GeneralSecurityException, IOExce
...
+   * . If `GOOGLE_API_USE_CLIENT_CERTIFICATE` environment variable is set to "true",
+   * the function looks for user provided client certificate first from 
+   * clientCertificateSource InputStream, if not exists, then the default from
+   * {@link Utils#loadDefaultCertificate()}. If client certificate exists,
+   * the transport uses it and is mutual TLS.
+   * 
+   * @param clientCertificateSource InputStream for mutual TLS client certificate and private key   
+   */
+  @Beta
+  public static NetHttpTransport newTrustedTransport(InputStream clientCertificateSource)
```
chingor13, line 78: Same comment as the ApacheTransport: we probably don't need to make this public, or alternatively provide the KeyStore/password.

google-api-client/src/main/java/com/google/api/client/googleapis/javanet/GoogleNetHttpTransport.java
```diff
@@ -51,10 +58,54 @@ static HttpTransport newProxyTransport() throws GeneralSecurityException, IOExce
...
+   * the function looks for user provided client certificate first from 
+   * clientCertificateSource InputStream, if not exists, then the default from
+   * {@link Utils#loadDefaultCertificate()}. If client certificate exists,
+   * the transport uses it and is mutual TLS. Note that mutual TLS may not work properly
+   * if you specify a proxy with the Builder instance.
+   * 
+   * @param clientCertificateSource InputStream for mutual TLS client certificate and private key   
+   */
+  @Beta
+  public static NetHttpTransport.Builder newTrustedTransportBuilder(
```
chingor13, line 98: Also probably not needed to be public.

google-api-client/src/main/java/com/google/api/client/googleapis/util/Utils.java
```diff
@@ -55,6 +71,82 @@ public static HttpTransport getDefaultTransport() {
...
 
+  /**
+   * Returns the `cert_provider_command` field in context_aware_metadata.json file.
+   * 
+   * @param contextAwareMetadataJson ~/.secureConnect/context_aware_metadata.json file content.
+   * @return `cert_provider_command` field
+   * @since 1.31
+   */
+  @SuppressWarnings("unchecked")
+  public static ArrayList<String> extractCertificateProviderCommand(String contextAwareMetadataJson) {
```
chingor13, line 82: Prefer the interface List over the implementation (ArrayList).

Also, this doesn't seem necessary to expose for developers.

```suggestion
  static List<String> extractCertificateProviderCommand(String contextAwareMetadataJson) {
```

google-api-client/src/main/java/com/google/api/client/googleapis/util/Utils.java
```diff
@@ -55,6 +71,82 @@ public static HttpTransport getDefaultTransport() {
...
+  }
+
+  /**
+   * Returns the default client certificate by running the cert_provider_command commands
+   * from ~/.secureConnect/context_aware_metadata.json file.
+   * 
+   * @return The default client certificate input stream
+   * @since 1.31
+   */
+  public static InputStream loadDefaultCertificate() throws IOException, GeneralSecurityException {
```
chingor13, line 95: Maybe `loadDefaultMtlsCertificate`? This is a general `Utils` class.

google-api-client/src/main/java/com/google/api/client/googleapis/util/Utils.java
chingor13: We know the structure/format of this file. We should use a strongly typed model for this. See google-http-client's GenericJson class: https://googleapis.github.io/google-http-java-client/json.html

We should also be parsing from an `InputStream` and not a `String`.

google-api-client/src/main/java/com/google/api/client/googleapis/util/Utils.java
```diff
@@ -55,6 +71,82 @@ public static HttpTransport getDefaultTransport() {
...
+    String json = new String(Files.readAllBytes(Paths.get(CONTEXT_AWARE_METADATA_PATH)));
+    ArrayList<String> command = extractCertificateProviderCommand(json);
+    
+    // Call the command.
+    Process process = new ProcessBuilder(command).start();
+    int exitCode = 0;
+    try {
+      exitCode = process.waitFor();
+    } catch (InterruptedException exception) {
+      throw new GeneralSecurityException("Failed to execute cert provider command", exception);
```
chingor13, line 111: This is probably not the right exception. `GeneralSecurityException` is for generally for crypto related security issues. Perhaps (`AccessControlException`)[https://docs.oracle.com/javase/7/docs/api/java/security/AccessControlException.html]?

chingor13 (changes requested):

chingor13:
BTW, google-http-client 1.38.0 is now available and updated in master

google-api-client/src/main/java/com/google/api/client/googleapis/util/Utils.java
arithmetic1728: changed to `AccessControlException`.

google-api-client/src/main/java/com/google/api/client/googleapis/util/Utils.java
arithmetic1728: done

google-api-client/src/main/java/com/google/api/client/googleapis/util/Utils.java
arithmetic1728: now used `GenericJson` and changed `String` to `InputStream`.

google-api-client/src/main/java/com/google/api/client/googleapis/javanet/GoogleNetHttpTransport.java
arithmetic1728: removed this method

google-api-client/src/main/java/com/google/api/client/googleapis/javanet/GoogleNetHttpTransport.java
arithmetic1728: now using keystore/password

google-api-client/pom.xml
arithmetic1728: removed gson dependency

google-api-client/src/main/java/com/google/api/client/googleapis/apache/v2/GoogleApacheHttpTransport.java
arithmetic1728: now using keystore/password

google-api-client/pom.xml
arithmetic1728: It seems there is no way to use Mockito to mock static for Java 7 since Java 7 doesn't support method reference and lambda, so I will still use PowerMockito.

google-api-client/src/test/java/com/google/api/client/googleapis/apache/v2/GoogleApacheHttpTransportTest.java
```diff
@@ -0,0 +1,71 @@
...
+    return SecurityUtils.createMtlsKeyStore(certAndKey);
+  }
+
+  // If client certificate shouldn't be used, then neither the provided mtlsKeyStore
+  // nor the default mtls key store should be used.
+  public void testNotUseCertificate() throws Exception {
+    KeyStore mtlsKeyStore = createTestMtlsKeyStore();
+    PowerMockito.spy(Utils.class);
+    PowerMockito.when(Utils.useMtlsClientCertificate()).thenReturn(false);
+    PowerMockito.when(Utils.loadDefaultMtlsKeyStore()).thenReturn(mtlsKeyStore);
```
chingor13, line 35: I think needing to mock static methods is a bit of a code-smell - we can wrap these helpers into a provider class which we can mock/inject for testing.

google-api-client/src/main/java/com/google/api/client/googleapis/util/Utils.java
```diff
@@ -55,6 +67,68 @@ public static HttpTransport getDefaultTransport() {
...
   }
+
+  /**
+   * Returns if mTLS client certificate should be used. mTLS client certificate is used if and only
+   * if "GOOGLE_API_USE_CLIENT_CERTIFICATE" environment variable value is "true".
+   *
+   * @return If mTLS client certificate should be used
+   * @since 1.31
+   */
+  public static Boolean useMtlsClientCertificate() {
```
chingor13, line 91: This is an implementation detail of the mTLS detection workflow and shouldn't need to be public.

chingor13 (changes requested):

google-api-client/src/main/java/com/google/api/client/googleapis/util/Utils.java
```diff
@@ -55,6 +67,68 @@ public static HttpTransport getDefaultTransport() {
...
+
+    // Load the cert provider command from the json file.
+    InputStream stream = new FileInputStream(file);
+    List<String> command = extractCertificateProviderCommand(stream);
+
+    // Call the command.
+    Process process = new ProcessBuilder(command).start();
+    int exitCode = 0;
+    try {
+      exitCode = process.waitFor();
```
chingor13, line 121: We should also set a timeout here otherwise this could potentially block the entire process.

google-api-client/src/main/java/com/google/api/client/googleapis/util/Utils.java
```diff
@@ -55,6 +67,68 @@ public static HttpTransport getDefaultTransport() {
...
+    } catch (InterruptedException exception) {
```
chingor13, line 122: This error should bubble up indicating that the process was interrupted.

google-api-client/src/main/java/com/google/api/client/googleapis/javanet/GoogleNetHttpTransport.java
```diff
@@ -51,10 +56,44 @@ static HttpTransport newProxyTransport() throws GeneralSecurityException, IOExce
...
+   * {@link Utils#loadDefaultMtlsKeyStore()} to create the transport. If either key
+   * store exists, then the created transport is mutual TLS. The provided key store
+   * takes precedence over the default one. 
+   * 
+   * @param mtlsKeyStore KeyStore for mutual TLS client certificate and private key
+   * @param mtlsKeyStorePassword KeyStore password
+   * @since 1.31   
+   */
+  @Beta
+  public static NetHttpTransport newTrustedTransport(KeyStore mtlsKeyStore, String mtlsKeyStorePassword)
```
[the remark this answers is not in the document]
arithmetic1728: This override is not intended to be general purpose. We need this override to control whether the keystore should be used for mtls, even if it is explicitly provided by the user. Python/Go has a similar override for the same functionality. We will probably need this control for a while until all mtls endpoints are GA. For googleapis, users should use this override in case they want their own certificate (even though this is discouraged).

google-api-client/src/main/java/com/google/api/client/googleapis/apache/v2/GoogleApacheHttpTransport.java
```diff
@@ -39,11 +40,33 @@
...
+   * to create the transport. If either key store exists, then the created transport is mutual TLS.
+   * The provided key store takes precedence over the default one.
+   *
+   * @param mtlsKeyStore KeyStore for mutual TLS client certificate and private key
+   * @param mtlsKeyStorePassword KeyStore password
+   */
+  @Beta
+  public static ApacheHttpTransport newTrustedTransport(
+      KeyStore mtlsKeyStore, String mtlsKeyStorePassword)
+      throws GeneralSecurityException, IOException {
```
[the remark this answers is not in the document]
arithmetic1728: same here. 

This override is not intended to be general purpose. We need this override to control whether the keystore should be used for mtls, even if it is explicitly provided by the user. Python/Go has a similar override for the same functionality. We will probably need this control for a while until all mtls endpoints are GA. For googleapis, users should use this override in case they want their own certificate (even though this is discouraged).

google-api-client/src/main/java/com/google/api/client/googleapis/util/MtlsUtils.java
```diff
@@ -0,0 +1,150 @@
...
+import com.google.common.annotations.VisibleForTesting;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.InputStream;
+import java.security.GeneralSecurityException;
+import java.security.KeyStore;
+import java.util.List;
+
+public class MtlsUtils {
```
chingor13, line 31: Since this necessitates being public (for use in `com.google.api.client.googleapis.javanet` and `com.google.api.client.googleapis.apache.v2`), we should consider flattening the public classes and perhaps putting in its own package (`com.google.api.client.googleapis.mtls` perhaps?).

google-api-client/src/test/java/com/google/api/client/googleapis/MtlsTransportBaseTest.java
```diff
@@ -0,0 +1,91 @@
...
+      return keyStore;
+    }
+  }
+
+  abstract protected HttpTransport buildTrustedTransport(MtlsUtils.MtlsProvider mtlsProvider) throws IOException, GeneralSecurityException;
+
+  // If client certificate shouldn't be used, then neither the provided mtlsKeyStore
+  // nor the default mtls key store should be used.
+  @Test
+  public void testNotUseCertificate() throws Exception {
```
chingor13, line 68: Not sure why I added the `throws Exception`, but theses should be more explicit

google-api-client/src/test/java/com/google/api/client/googleapis/MtlsTransportBaseTest.java
```diff
@@ -0,0 +1,91 @@
...
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.security.GeneralSecurityException;
+import java.security.KeyStore;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+public abstract class MtlsTransportBaseTest  {
```
chingor13, line 30: We'll want to add more tests for the failure cases (when loadDefaultKeyStore() can fail) and how the `newTrustedTransport()` behaves in those cases.

google-api-client/src/main/java/com/google/api/client/googleapis/util/MtlsUtils.java
```diff
@@ -0,0 +1,150 @@
...
+  public interface MtlsProvider {
+    boolean useMtlsClientCertificate();
+
+    String getKeyStorePassword();
+
+    KeyStore loadDefaultKeyStore() throws IOException, GeneralSecurityException;
```
chingor13, line 37: This should probably be renamed `getKeyStore()`. The `DefaultMtlsProvider#getKeyStore()` is the "loadDefaultKeyStore()"

google-api-client/src/test/java/com/google/api/client/googleapis/MtlsTransportBaseTest.java
chingor13: We can probably also add a test for the parameterless `newTrustedTransport` and use [`Assume.assumeTrue`](https://junit.org/junit4/javadoc/4.12/org/junit/Assume.html#assumeTrue(boolean)) to protected the tests (skip if the test is not run under certain conditions) and we can possibly set up additional CI runs that configure mTLS on that test runner.

chingor13 (changes requested):

google-api-client/src/main/java/com/google/api/client/googleapis/javanet/GoogleNetHttpTransport.java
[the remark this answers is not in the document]
chingor13: If this is necessary, with the refactor to an `MtlsProvider`, we can provide a custom provider where the user can provide their own `KeyStore` and make the `newTrustedTransport(MtlsProvider)` public.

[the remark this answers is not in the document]
arithmetic1728: Now I think it, it should be fine just using the default cert.

google-api-client/src/main/java/com/google/api/client/googleapis/util/Utils.java
arithmetic1728: Timeout is added.

google-api-client/src/main/java/com/google/api/client/googleapis/util/MtlsUtils.java
arithmetic1728: yes, created a new `mtls` package

google-api-client/src/test/java/com/google/api/client/googleapis/MtlsTransportBaseTest.java
arithmetic1728: done

google-api-client/src/main/java/com/google/api/client/googleapis/util/MtlsUtils.java
arithmetic1728: done

google-api-client/src/test/java/com/google/api/client/googleapis/MtlsTransportBaseTest.java
arithmetic1728: added test for `newTrustedTransport` where `loadKeyStore` fails.

The parameterless `newTrustedTransport` tests with CI configuration I think we can add it later.

google-api-client/src/main/java/com/google/api/client/googleapis/mtls/MtlsUtils.java
```diff
@@ -0,0 +1,162 @@
...
+
+    @Override
+    public String getKeyStorePassword() {
+      return "";
+    }
+
+    @Override
+    public KeyStore getKeyStore() throws IOException, GeneralSecurityException {
+      // Load the cert provider command from the json file.
+      InputStream stream;
```
elharo, line 85: make the try block bigger so you don't have to split the variable declaration and initialization

google-api-client/src/main/java/com/google/api/client/googleapis/mtls/MtlsUtils.java
```diff
@@ -0,0 +1,162 @@
...
+    @VisibleForTesting
+    static int runCertificateProviderCommand(Process commandProcess, long timeoutMilliseconds)
+        throws IOException, InterruptedException {
+      long startTime = System.currentTimeMillis();
+      long remainTime = timeoutMilliseconds;
+      boolean terminated = false;
+
+      do {
+        try {
+          // Check if process is terminated by pooling the exitValue, which throws
```
elharo, line 130: pooling --> polling

google-api-client/src/main/java/com/google/api/client/googleapis/mtls/MtlsUtils.java
```diff
@@ -0,0 +1,162 @@
...
+      return commandProcess.exitValue();
+    }
+  }
+
+  private static final MtlsProvider MTLS_PROVIDER = new DefaultMtlsProvider();
+
+  /**
+   * Returns the default MtlsProvider instance.
+   *
+   * @return The default MtlsProvider instance.
```
elharo, line 157: nit: "the default MtlsProvider instance" per javadoc guidelines
no period

google-api-client/src/main/java/com/google/api/client/googleapis/mtls/MtlsUtils.java
arithmetic1728: done

google-api-client/src/main/java/com/google/api/client/googleapis/mtls/MtlsUtils.java
arithmetic1728: done

google-api-client/src/main/java/com/google/api/client/googleapis/mtls/MtlsUtils.java
arithmetic1728: The current way seems fine since it catches exception right after each call so we know which function throws which exception. A bigger try block would basically wrap the whole implementation.

elharo: One big try block that wraps all the code is exactly what how exceptions in Java were designed to be used. In this method I see no ambiguity about which lines throw which exceptions.

elharo (changes requested):

google-api-client/src/main/java/com/google/api/client/googleapis/mtls/MtlsUtils.java
arithmetic1728: I see. Changed to one big try block.

google-api-client/src/main/java/com/google/api/client/googleapis/mtls/MtlsUtils.java
```diff
@@ -81,33 +81,26 @@ public String getKeyStorePassword() {
...
-        stream = new FileInputStream(metadataPath);
+        // Load the cert provider command from the json file.
+        InputStream stream = new FileInputStream(metadataPath);
+        List<String> command = extractCertificateProviderCommand(stream);
+
+        // Run the command and timeout after 1000 milliseconds.
+        Process process = new ProcessBuilder(command).start();
+        int exitCode = runCertificateProviderCommand(process, 1000);
+        if (exitCode != 0) {
+          throw new IOException(String.format("Failed to execute cert provider command with exit code: %d", exitCode));
```
elharo, line 93: simple string concatenation would be clearer

elharo (approved):

google-api-client/src/main/java/com/google/api/client/googleapis/mtls/MtlsUtils.java
arithmetic1728: done

google-api-client/src/main/java/com/google/api/client/googleapis/mtls/package-info.java
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright 2020 Google Inc.
```
chingor13, line 2: nit:
```suggestion
 * Copyright 2020 Google LLC
```

chingor13 (changes requested):
One tiny license header nit, otherwise LGTM.

chingor13 (approved):

PR #1637, opened 2020-11-13
fix: socketfactory not registered for apache

arithmetic1728:
For `GoogleApacheHttpTransport`, `socketFactory` has to be registered for https, and provided to `connectionManager`. This PR does the following:

- register socket factory for `http` and `https`, and pass to connection manager ([reference](https://github.com/mydevotion/httpclient/blob/master/httpclient/src/main/java/org/apache/http/impl/conn/PoolingHttpClientConnectionManager.java#L109))
- removed `.setSSLSocketFactory(socketFactory)` from the chain of calls after `HttpClientBuilder.create()`, since it will be overwritten by `.setConnectionManager(connectionManager)`


elharo (commented):
This could use a test.

arithmetic1728:
> This could use a test.

Refactored the code and added test.

chingor13 (approved):

karatalaya:
- [ ] [[@****]()]()

PR #1661, opened 2021-01-06
fix: use Gson instead of Jackson

elharo:
@BenWhitehead


elharo (commented):
Ping ping ping

chingor13 (approved):

hiranya911:
@chingor13 @elharo 

Can you share some context on why this change was made? This change is making it difficult for us (Firebase) to upgrade our SDK dependency from Google API client 1.30.x to 1.31.x. The use of Gson significantly changes how certain JSON input is handled, which is causing numerous test failures in our code.

elharo:
go/json-madness

hiranya911:
Thanks @elharo. I've submitted a detailed bug report with our findings at #1779

PR #1665, opened 2021-01-12
fix: switch to use google-http-client-gson by default

chingor13:


elharo (changes requested):
need to fix tests too

elharo (commented):
dupe of #1661

PR #1792, opened 2021-05-10
fix: add the parent field

Neenu1995:
Adding the parent field back. It is a required field for release.


elharo:
What's the error message you got? It's possible the parent pom is supplying something that's needed but the parent itself, and this parent pom in particular, shouldn't be required.

PR #1960, opened 2022-01-19
chore: downstream check for all libraries in single job

Neenu1995:


.kokoro/downstream-client-library-check.sh
```diff
@@ -0,0 +1,115 @@
...
+
+mkdir /tmp/foo && cd /tmp/foo
+wget https://dl.google.com/dl/android/maven2/com/google/android/gms/play-services-basement/8.3.0/play-services-basement-8.3.0.aar
+unzip play-services-basement-8.3.0.aar
+mvn install:install-file \
+  -Dfile=classes.jar \
+  -DgroupId=com.google.android.google-play-services \
+  -DartifactId=google-play-services \
+  -Dversion=1 \
+  -Dpackaging=jar
```
lesv, line 34: You might wish to explain why you need this.

lesv (approved):
one comment, otherwise LGTM

Also, please mention in both files that "there may be individual failures that you may choose not to block on if it's known they have problems at the moment".

lesv (approved):

.kokoro/downstream-client-library-check.sh
Neenu1995: Done

PR #2044, opened 2022-04-19
fix: Error details improvement - Approach 2 - Remove Details field

blakeli0:
This approach removes `Details` field from `GoogleJsonError`, because `GoogleJsonError` is already extending `GenericJson` it should be able to handle any arbitrary response as well. However, this is a breaking change.


blakeli0:
Closing in favor of https://github.com/googleapis/google-api-java-client/pull/2210

PR #2069, opened 2022-05-10
chore: set Java header during native image runtime

mpeddada1:
This contributes towards removing the use of the SVM dependency for native image compilation. This code makes use of the org.graalvm.nativeimage.imagecode system property instead of `Substitutions` to set the Java version at image runtime. For reference: https://www.graalvm.org/sdk/javadoc/index.html?constant-values.html

See reproducer: https://github.com/mpeddada1/avoid-substitutions

### DIFF
--- a/google-api-client/src/main/java/com/google/api/client/googleapis/services/AbstractGoogleClientRequest.java
+++ b/google-api-client/src/main/java/com/google/api/client/googleapis/services/AbstractGoogleClientRequest.java
@@ -37,6 +37,7 @@ import com.google.api.client.http.HttpResponseInterceptor;
 import com.google.api.client.http.UriTemplate;
 import com.google.api.client.util.GenericData;
 import com.google.api.client.util.Preconditions;
+import com.google.common.base.Joiner;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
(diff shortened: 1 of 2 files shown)


google-api-client/src/main/java/com/google/api/client/googleapis/services/AbstractGoogleClientRequest.java
```diff
@@ -164,6 +165,16 @@ static class ApiClientVersion {
     }
 
     public String toString() {
+      // When running the application as a native image, append `-graalvm` to the
+      // version.
+      String imageCode = System.getProperty("org.graalvm.nativeimage.imagecode");
```
meltsufin, line 170: Please use constants instead of string literals here.

meltsufin (approved):

google-api-client/src/main/java/com/google/api/client/googleapis/services/AbstractGoogleClientRequest.java
mpeddada1: That's a valid point! However, this might come down to the question of whether it is worth having dependency on graal-sdk in this repo? If we use the latest version (22.1.0) then we'll see the following incompatibility: 
```
bad class file: /home/runner/.m2/repository/org/graalvm/sdk/graal-sdk/22.1.0/graal-sdk-22.1.0.jar(org/graalvm/nativeimage/ImageInfo.class)
Error:      class file has wrong version 55.0, should be 52.0
Error:      Please remove or make sure it appears in the correct subdirectory of the classpath.
Error:  -> [Help 1]
```
In which case, we may need to do something similar to https://github.com/googleapis/gax-java/pull/1671. This could still pose a slight risk since we still target Java 7. But open to hearing about what you think!

meltsufin: I'm fine with either. @suztomo what do you think? Is it worth it to bring the whole setup for compiling with 11 and testing on 7 in this case?

PR #2109, opened 2022-07-14
fix: add error description to batch emptiness validation

BenWhitehead:
When invoking BatchRequest#execute(), we ensure the batch contains at
least one request to process. Previously this check, did not provide any
error message. Now, if the check fails "Batch is empty" will be the
message of the resulting exception.

Related to https://github.com/googleapis/java-storage/issues/694


Neenu1995 (approved):

TimurSadykov (approved):
LGTM

PR #2195, opened 2022-11-29
fix: update and declare commons-codec dependency

alicejli:
This PR manually declares the `commons-codec` version to `1.15`. 

It was discovered that `commons-codec-1.11.jar` has a    a medium vulnerability (https://sca.analysiscenter.veracode.com/vulnerability-database/security/sca/vulnerability/sid-22742/summary). This vulnerability does not have a CVE and it is related to Insecure Input Validation and encoding. 

This is an indirect dpendency that the API client library uses indirectly by using Apache HttpClient.
Any version of codec v1.13+ should have the fix for this vulnerability.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -203,6 +203,11 @@
         <artifactId>google-api-client-xml</artifactId>
         <version>${project.version}</version>
       </dependency>
+      <dependency>
+        <groupId>commons-codec</groupId>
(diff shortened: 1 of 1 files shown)


suztomo (approved):
Approving. Can you add comment about the change?

Issue #1974, opened 2022-01-28
GMB API error model does not match GoogleJsonError model


ghost:
The error response model of the GMB APIs ([https://developers.google.com/my-business/ref_overview](https://developers.google.com/my-business/ref_overview)) currently looks like this:

```
{
    "error": {
        "code": 400,
        "message": "Request contains an invalid argument.",
        "status": "INVALID_ARGUMENT",
        "details": [{
                "@type": "type.googleapis.com/google.mybusiness.v4.ValidationError",
                "errorDetails": [{
                        "code": 3,
                        "field": "regular_hours.periods.close_time",
                        "message": "Time field must follow hh:mm format.",
                        "value": "25:00"
                    }
                ]
            }
        ]
    }
}
```

The subclass [Details](https://github.com/googleapis/google-api-java-client/blob/ef26931d0b3392d7e4354d22d38c1f95fcd9c6a0/google-api-client/src/main/java/com/google/api/client/googleapis/json/GoogleJsonError.java#L187) of the [GoogleJsonError](https://github.com/googleapis/google-api-java-client/blob/main/google-api-client/src/main/java/com/google/api/client/googleapis/json/GoogleJsonError.java) class, however, does not feature an errorDetails property.
Once the [JsonParser](https://github.com/googleapis/google-http-java-client/blob/main/google-http-client/src/main/java/com/google/api/client/json/JsonParser.java) is reached, we thus reach [ customizeParser.handleUnrecognizedKey(destination, key);](https://github.com/googleapis/google-http-java-client/blob/f7948014150d0205ffb6abf322e97e0c03b4ee16/google-http-client/src/main/java/com/google/api/client/json/JsonParser.java#L462) for the errorDetails field.

lesv:
@danielgazineu This is an issue w/ APIary that needs some love.

blakeli0:
@vpriesner Can you please provide the exact reproducible sequence for us to troubleshoot this issue?

PR #2210, opened 2022-12-16
fix: Make Details field extends GenericJson so that GoogleJsonError should include any arbitrary error info from services

blakeli0:
fixes: #1974 
The root cause of the issue is that `details` attribute from backend could be of any arbitrary type, we need a generic way to parse error details from backend. 
This approach makes `Details` field extends `GenericJson` so that it can parse any data from backend. This is the preferred approach since this is the simplest and most generic one.
Other approaches: #2044, #2211


meltsufin (commented):
As discussed offline, I like this approach because it's the most future-proof. Let's polish up the tests, and then it's good to go.

meltsufin (approved):

Skrilla84 (changes requested):
Compromised accounts, taking action asap

PR #2211, opened 2022-12-16
fix: Error details improvement - Approach 3 - Add known types to Details

blakeli0:
This approach tries to add all known types(not done yet, just tries to illustrate the idea) to `Details` field, similar to what we did for GAPIC. 
However, unlike GAPIC where we can rely on an [error_details.proto](https://github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto) for vast majority of standard error types, seems like errors returned from Apiary client library have much more varieties and a lot of them are not standard errors(The three listed types `ParameterViolations`, `FieldViolations`, `ModificationErrors` are all non-standard). This makes this approach almost infeasible.


blakeli0:
Closing in favor of #2210

PR #2242, opened 2023-01-25
feat: remove deprecated OOB flow

meltsufin:
Also remove `google-api-client-java6` module which is empty after removal of `GooglePromptReceiver.java`.


google-api-client/src/main/java/com/google/api/client/googleapis/auth/oauth2/GoogleOAuthConstants.java
```diff
@@ -40,12 +40,5 @@ public class GoogleOAuthConstants {
   public static final String DEFAULT_PUBLIC_CERTS_ENCODED_URL =
       "https://www.googleapis.com/oauth2/v1/certs";
 
-  /**
-   * Redirect URI to use for an installed application as specified in <a
-   * href="https://developers.google.com/identity/protocols/OAuth2InstalledApp">Using OAuth 2.0 for
-   * Mobile & Desktop Apps</a>.
-   */
-  public static final String OOB_REDIRECT_URI = "urn:ietf:wg:oauth:2.0:oob";
```
meltsufin, line 48: clirr is complaining about removal of this constant. It wasn't deprecated before. So, is it OK to remove?

Neenu1995: Since the OOB flow will be blocked for all clients on January 31, I think it is safe to remove it. Keeping it won't prevent existing users from being broken.
That is my take, I will eave the final decision to you.

Neenu1995 (approved):

google-api-client/src/main/java/com/google/api/client/googleapis/auth/oauth2/GoogleOAuthConstants.java
suztomo: I agree with Neenu.

Separately, can we hav one more release with "@Deprecated" annotation with a link to a document that explains alternative(s).

suztomo: Can we release this documentation enhancement https://github.com/googleapis/google-api-java-client/pull/2244 first before stop publishing the artifact?

PR #2435, opened 2024-02-20
feat: Add Universe Domain Support

lqiu96:
Universe Domain support can be done almost all inside this PR. There is an additional change required inside the google-api-client-services generator. Need to add the setter inside the `{Service}.Builder` implementation.
```
@Override
public Builder setUniverseDomain(String universeDomain) {
  return (Builder) super.setUniverseDomain(universeDomain);
}
```

The reason this needs to be in the generated client code is because it returns the parent's implementation returns a type of `AbstractGoogleClient.Builder` instead of `Bigquery.Builder`. Having this inside the child implementation would return the correct type (`Bigquery.Builder`).

Tested this locally with a local build of Bigquery Apiary client and a local build of the api-client library. Connection to the test environment works and the validation is called on each RPC invocation.

### Changes required in Java-Storage and Java-Bigquery
```
{ApiaryClient}.Builder()
            .setRootUrl(options.getResolvedHost("{Apiary Client Service Name}"))
            .setUniverseDomain(options.getUniverseDomain())
.build()
```
1. Setting the rootUrl will now use the resolvedHost and no longer use the apiary workaround
2. Pass the universe domain in to the apiary client (even though setting the rootUrl will be the rootUrl used and the universe domain configuration value isn't used in resolving the endpoint, pass in the universe domain as this will allow users to call the universe domain getter).

### Local Tests
1. Local Bigquery Apiary - Able to connect + Universe Domain Validation
3. Local Java-Bigquery client library - Able to connect + Universe Domain Validation

### Next Steps:
PR in google-api-java-client-services: https://github.com/googleapis/google-api-java-client-services/pull/19934
- Requires merging in this PR and a release of a new feat (minor version bump).

### DIFF
--- a/google-api-client/src/main/java/com/google/api/client/googleapis/services/AbstractGoogleClient.java
+++ b/google-api-client/src/main/java/com/google/api/client/googleapis/services/AbstractGoogleClient.java
@@ -25,6 +25,7 @@ import com.google.auth.http.HttpCredentialsAdapter;
 import com.google.common.annotations.VisibleForTesting;
 import java.io.IOException;
 import java.util.logging.Logger;
+import java.util.regex.Pattern;
 
 /**
  * Abstract thread-safe Google client.
@@ -406,6 +407,13 @@ public abstract class AbstractGoogleClient {
     String universeDomain;
 
     /**
+     * Regex pattern to check if the URL passed in matches the default endpoint confgured from a
+     * discovery doc. Follows the format of `https://{serviceName}(.mtls).googleapis.com/`
+     */
+    Pattern defaultEndpointRegex =
+        Pattern.compile("https://[a-zA-Z]*(\\.mtls)?\\.googleapis.com/?");
+
+    /**
      * Whether the user has configured an endpoint via {@link #setRootUrl(String)}. This is added in
      * because the rootUrl is set in the Builder's constructor. ,
      *
@@ -443,8 +451,7 @@ public abstract class AbstractGoogleClient {
       this.servicePath = normalizeServicePath(servicePath);
       this.httpRequestInitializer = httpRequestInitializer;
       this.serviceName = parseServiceName(rootUrl);
-      this.isUserConfiguredEndpoint =
-          !this.rootUrl.endsWith(Credentials.GOOGLE_DEFAULT_UNIVERSE + "/");
+      this.isUserConfiguredEndpoint = !defaultEndpointRegex.matcher(this.rootUrl).matches();
     }
 
     /**
--- a/google-api-client/src/main/java/com/google/api/client/googleapis/services/AbstractGoogleClient.java
+++ b/google-api-client/src/main/java/com/google/api/client/googleapis/services/AbstractGoogleClient.java
@@ -154,12 +154,12 @@ public abstract class AbstractGoogleClient {
    *     Domain in the Credentials
    */
   public void validateUniverseDomain() throws IOException {
-    String expectedUniverseDomain;
-    if (!(httpRequestInitializer instanceof HttpCredentialsAdapter)) {
-      expectedUniverseDomain = Credentials.GOOGLE_DEFAULT_UNIVERSE;
-    } else {
+    String expectedUniverseDomain = Credentials.GOOGLE_DEFAULT_UNIVERSE;
+    if (httpRequestInitializer instanceof HttpCredentialsAdapter) {
       Credentials credentials = ((HttpCredentialsAdapter) httpRequestInitializer).getCredentials();
-      expectedUniverseDomain = credentials.getUniverseDomain();
+      if (credentials != null) {
+        expectedUniverseDomain = credentials.getUniverseDomain();
+      }
     }
     if (!expectedUniverseDomain.equals(getUniverseDomain())) {
       throw new IllegalStateException(
@@ -409,7 +409,7 @@ public abstract class AbstractGoogleClient {
     String universeDomain;
 
     /**
-     * Regex pattern to check if the URL passed in matches the default endpoint confgured from a
+     * Regex pattern to check if the URL passed in matches the default endpoint configured from a
      * discovery doc. Follows the format of `https://{serviceName}(.mtls).googleapis.com/`
      */
     Pattern defaultEndpointRegex =
@@ -424,8 +424,8 @@ public abstract class AbstractGoogleClient {
      * done via {@link #setRootUrl(String)}.
      *
      * <p>For other uses cases that touch this Builder's constructor directly, check if the rootUrl
-     * passed in references the Google Default Universe (GDU). Any rootUrl value that is not set in
-     * the GDU is a user configured endpoint.
+     * passed matches the default endpoint regex. If it doesn't match, it is a user configured
+     * endpoint.
      */
     boolean isUserConfiguredEndpoint;
 
--- a/google-api-client/src/main/java/com/google/api/client/googleapis/services/AbstractGoogleClient.java
+++ b/google-api-client/src/main/java/com/google/api/client/googleapis/services/AbstractGoogleClient.java
@@ -145,8 +145,7 @@ public abstract class AbstractGoogleClient {
    * uses the HttpRequestInitializer to get the Credentials and is enforced that the
    * HttpRequestInitializer is of the {@see <a
    * href="https://github.com/googleapis/google-auth-library-java/blob/main/oauth2_http/java/com/google/auth/http/HttpCredentialsAdapter.java">HttpCredentialsAdapter</a>}
-   * from the google-auth-library. If the HttpCredentialsAdapter is not used, the configured
-   * Universe Domain is validated against the Google Default Universe (GDU): `googleapis.com`.
+   * from the google-auth-library.
    *
    * <p>To use a non-GDU Credentials, you must use the HttpCredentialsAdapter class.
    *
@@ -155,20 +154,19 @@ public abstract class AbstractGoogleClient {
    *     Domain in the Credentials
    */
   public void validateUniverseDomain() throws IOException {
-    String expectedUniverseDomain = Credentials.GOOGLE_DEFAULT_UNIVERSE;
     if (httpRequestInitializer instanceof HttpCredentialsAdapter) {
       Credentials credentials = ((HttpCredentialsAdapter) httpRequestInitializer).getCredentials();
-      // No need for a null check as HttpCredentialsAdapter cannot be initialized will null
+      // No need for a null check as HttpCredentialsAdapter cannot be initialized with null
       // Credentials
-      expectedUniverseDomain = credentials.getUniverseDomain();
-    }
-    if (!expectedUniverseDomain.equals(getUniverseDomain())) {
-      throw new IllegalStateException(
-          String.format(
-              "The configured universe domain (%s) does not match the universe domain found"
-                  + " in the credentials (%s). If you haven't configured the universe domain"
-                  + " explicitly, `googleapis.com` is the default.",
-              getUniverseDomain(), expectedUniverseDomain));
+      String expectedUniverseDomain = credentials.getUniverseDomain();
+      if (!expectedUniverseDomain.equals(getUniverseDomain())) {
+        throw new IllegalStateException(
+            String.format(
+                "The configured universe domain (%s) does not match the universe domain found"
+                    + " in the credentials (%s). If you haven't configured the universe domain"
+                    + " explicitly, `googleapis.com` is the default.",
+                getUniverseDomain(), expectedUniverseDomain));
+      }
     }
   }
 
--- a/google-api-client/src/main/java/com/google/api/client/googleapis/services/AbstractGoogleClient.java
+++ b/google-api-client/src/main/java/com/google/api/client/googleapis/services/AbstractGoogleClient.java
@@ -154,19 +154,20 @@ public abstract class AbstractGoogleClient {
    *     Domain in the Credentials
    */
   public void validateUniverseDomain() throws IOException {
-    if (httpRequestInitializer instanceof HttpCredentialsAdapter) {
-      Credentials credentials = ((HttpCredentialsAdapter) httpRequestInitializer).getCredentials();
-      // No need for a null check as HttpCredentialsAdapter cannot be initialized with null
-      // Credentials
-      String expectedUniverseDomain = credentials.getUniverseDomain();
-      if (!expectedUniverseDomain.equals(getUniverseDomain())) {
-        throw new IllegalStateException(
-            String.format(
-                "The configured universe domain (%s) does not match the universe domain found"
-                    + " in the credentials (%s). If you haven't configured the universe domain"
-                    + " explicitly, `googleapis.com` is the default.",
-                getUniverseDomain(), expectedUniverseDomain));
-      }
+    if (!(httpRequestInitializer instanceof HttpCredentialsAdapter)) {
+      return;
+    }
+    Credentials credentials = ((HttpCredentialsAdapter) httpRequestInitializer).getCredentials();
+    // No need for a null check as HttpCredentialsAdapter cannot be initialized with null
+    // Credentials
+    String expectedUniverseDomain = credentials.getUniverseDomain();
+    if (!expectedUniverseDomain.equals(getUniverseDomain())) {
+      throw new IllegalStateException(
+          String.format(
+              "The configured universe domain (%s) does not match the universe domain found"
+                  + " in the credentials (%s). If you haven't configured the universe domain"
+                  + " explicitly, `googleapis.com` is the default.",
+              getUniverseDomain(), expectedUniverseDomain));
     }
   }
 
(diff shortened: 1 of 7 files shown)


google-api-client/src/main/java/com/google/api/client/googleapis/services/AbstractGoogleClient.java
```diff
@@ -328,9 +439,35 @@ protected Builder(
         HttpRequestInitializer httpRequestInitializer) {
       this.transport = Preconditions.checkNotNull(transport);
       this.objectParser = objectParser;
-      setRootUrl(rootUrl);
-      setServicePath(servicePath);
+      this.rootUrl = normalizeRootUrl(rootUrl);
+      this.servicePath = normalizeServicePath(servicePath);
       this.httpRequestInitializer = httpRequestInitializer;
+      this.serviceName = parseServiceName(rootUrl);
+      this.isUserConfiguredEndpoint =
```
blakeli0, line 446: Why do we need to determine if the `rootUrl` is user configured based on if it ends with GDU in the `Builder`? Is setting `this.isUserConfiguredEndpoint = true;` in the setter not enough ?
Is it for the scenarios that customers may extend this class and call this Builder directly? Because I think in the public client, `rootUrl` is hidden. e.g. in [Bigquery](https://github.com/googleapis/google-api-java-client-services/blob/a3b11763afceda22d5b82b7550bf8d74617464b9/clients/google-api-services-bigquery/v2/2.0.0/com/google/api/services/bigquery/Bigquery.java#L8012-L8013)

google-api-client/src/main/java/com/google/api/client/googleapis/services/AbstractGoogleClient.java
```diff
@@ -311,6 +402,26 @@ public abstract static class Builder {
...
+    /**
+     * Whether the user has configured an endpoint via {@link #setRootUrl(String)}. This is added in
+     * because the rootUrl is set in the Builder's constructor. ,
+     *
+     * <p>Apiary clients don't allow user configurations to this Builder's constructor, so this
+     * would be set to false by default for Apiary libraries. User configuration to the rootUrl is
+     * done via {@link #setRootUrl(String)}.
+     *
+     * <p>For other uses cases that touch this Builder's constructor directly, check if the rootUrl
+     * passed in references the Google Default Universe (GDU). Any rootUrl value that is not set in
```
blakeli0, line 417: I don't think this logic is correct, customers could set regional endpoints which also end with GDU, see https://cloud.google.com/storage/docs/regional-endpoints

blakeli0:
> ### Changes required in Java-Storage and Java-Bigquery
> ```
> {ApiaryClient}.Builder()
>             .setRootUrl(options.getResolvedHost("{Apiary Client Service Name}"))
>             .setUniverseDomain(options.getUniverseDomain())
> .build()
> ```
> 
> 1. Setting the rootUrl will now use the resolvedHost and no longer use the apiary workaround
> 2. Pass the universe domain in to the apiary client (even though setting the rootUrl will be the rootUrl used and the universe domain configuration value isn't used in resolving the endpoint, pass in the universe domain as this will allow users to call the universe domain getter).

Why they have to set both rootUrl and universeDomain? I think setting universeDomain alone should be good enough?

google-api-client/src/main/java/com/google/api/client/googleapis/services/AbstractGoogleClient.java
lqiu96: > Is it for the scenarios that customers may extend this class and call this Builder directly? 

Yep, it's for the (hopefully) rare user use-case where users are directly extending this class and calling the builder. The apiary libraries don't expose the endpoint configurations in the Builder directly.

google-api-client/src/main/java/com/google/api/client/googleapis/services/AbstractGoogleClient.java
lqiu96: Ah. Turns out _I think_ this works, but for the wrong reasons. I believe we're parsing out the serviceName and then reconstructing it to build the the same endpoint later on. It's marked as a non-user configured endpoint.

I think it might be better if I change the `isUserConfiguredEndpoint` logic to match for a strict regex instead to cover the regional endpoints edgecase.

lqiu96:
> > ### Changes required in Java-Storage and Java-Bigquery
> > ```
> > {ApiaryClient}.Builder()
> >             .setRootUrl(options.getResolvedHost("{Apiary Client Service Name}"))
> >             .setUniverseDomain(options.getUniverseDomain())
> > .build()
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 1. Setting the rootUrl will now use the resolvedHost and no longer use the apiary workaround
> > 2. Pass the universe domain in to the apiary client (even though setting the rootUrl will be the rootUrl used and the universe domain configuration value isn't used in resolving the endpoint, pass in the universe domain as this will allow users to call the universe domain getter).
> 
> Why they have to set both rootUrl and universeDomain? I think setting universeDomain alone should be good enough?

The apiary wrapped libraries (java-storage and java-bigquery) may have some user configurations where they change the rootUrl value there. We'll need those configurations so that we won't override them when we resolve the endpoint here.

blakeli0:
> > > ### Changes required in Java-Storage and Java-Bigquery
> > > ```
> > > {ApiaryClient}.Builder()
> > >             .setRootUrl(options.getResolvedHost("{Apiary Client Service Name}"))
> > >             .setUniverseDomain(options.getUniverseDomain())
> > > .build()
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > 1. Setting the rootUrl will now use the resolvedHost and no longer use the apiary workaround
> > > 2. Pass the universe domain in to the apiary client (even though setting the rootUrl will be the rootUrl used and the universe domain configuration value isn't used in resolving the endpoint, pass in the universe domain as this will allow users to call the universe domain getter).
> > 
> > 
> > Why they have to set both rootUrl and universeDomain? I think setting universeDomain alone should be good enough?
> 
> The apiary wrapped libraries (java-storage and java-bigquery) may have some user configurations where they change the rootUrl value via settings there (for emulators, local endpoint, etc). We'll need those configurations so that we won't override them when the apiary library tries to resolve the endpoint in the code here.

I see. If that's the case, I would say we don't have to pass `universeDomain`, setting the `rootUrl` should be good enough. Because 
1. The Apiary client is hidden from customers in Bigquery/Storage, even if customers need to, they would use `ServiceOptions.getUniverseDomain()`.
2. This is less confusing to Bigquery/Storage developers and customers. As they may think it is required to pass both values.
3. This requires no change in Bigquery/Storage.

lqiu96:
I believe you're right regarding resolving the host. I also just realized that the universe domain is also needed as we need it to validate the universe domain value. If not passed, the resolved universe domain will default to `googleapis.com`. Passing in the configured universe domain value will use that for validation.

blakeli0:
> Previously, the handwritten libraries were validating the universe domain, but since we're moving it the apiary client, I believe we'll need to pass the configured universe domain value here.

It maybe a scenario we missed before, do we still want to validate the universe domain if the endpoint/rootUrl is provided? I think we should not because the endpoint configuration would override the universe domain configuration if both are configured. Otherwise it would be confusing if we pass both to the client, use one for validation but use another one for making calls.

google-api-client/src/main/java/com/google/api/client/googleapis/services/AbstractGoogleClient.java
```diff
@@ -328,9 +447,34 @@ protected Builder(
...
+
+    /**
+     * This is intended to invoked once on the initial Builder's constructor call. This parses the
+     * rootUrl value (set from the Discovery Doc) to use for the serviceName. The serviceName is
+     * used to construct an endpoint when the user passes in a custom Universe Domain value.
+     *
+     * <p>The roolUrl from the Discovery Docs will always follow the format of
+     * https://{serviceName}(.mtls).googleapis.com/
+     */
+    private String parseServiceName(String rootUrl) {
```
blakeli0, line 465: I think we can use the regex to extract the service name now.

lqiu96:
> > Previously, the handwritten libraries were validating the universe domain, but since we're moving it the apiary client, I believe we'll need to pass the configured universe domain value here.
> 
> It maybe a scenario we missed before, do we still want to validate the universe domain if the endpoint/rootUrl is provided? I think we should not because the endpoint configuration would override the universe domain configuration if both are configured. Otherwise it would be confusing if we pass both to the client, use one for validation but use another one for making calls.

It is a bit of weird scenario especially given Apiary's non-normal use case. I think we _should_ be validating the universe domain even if we are providing the "correct" endpoint from ServiceOptions.

Off the top of my head, I can think of these scenarios in which I think validation is needed required: 
1. User sets custom rootUrl (to the correct non-GDU endpoint) and no universeDomain. This will create a connection to the right place and the universe domain is resolved to GDU. They supply the correct Credentials that points to non-GDU. While this does _technically_ work as the Credentials is valid, this does not share the same functionality with the other client libs. GAPICs would throw an exception that the explicitly configured universe domain does not match before the call goes through.
2. User sets custom rootUrl (to the correct non-GDU endpoint) and a universeDomain. However, they supply the incorrect Credentials that points to GDU. I believe this would fail once the RPC hits the server, but the call should not go through (it should fail during validation). GAPICs would throw an exception that the explicitly configured universe domain does not match.

> Otherwise it would be confusing if we pass both to the client, use one for validation but use another one for making calls.

Agreed. Ideally the endpoint should be resolved inside the Apiary client itself for every use case. It's only a special case for these two handwritten apiary wrappers as Java-Core is used to resolved the endpoint. 

I chose this path because Java-Core sets the the host value to be the value of DEFUALT_HOST (https://github.com/googleapis/sdk-platform-java/blob/4b44a7851dc1d4fd2ac21a54df6c24db5625223c/java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java#L86) if nothing is configured. One option that I thought about would be to have the two libraries do some additional logic to filter this out. 
i.e.
```
String host = options.getHost().equals(DEFAULT_HOST) ? null : options.getHost();
{ApiaryClient}.Builder()
            .setRootUrl(host)
            .setUniverseDomain(options.getUniverseDomain())
.build()
```

I didn't go with this path since it's additional work for the two handwritten libraries.

google-api-client/src/main/java/com/google/api/client/googleapis/services/AbstractGoogleClient.java
```diff
@@ -88,6 +99,74 @@ protected AbstractGoogleClient(Builder builder) {
...
+  public void validateUniverseDomain() throws IOException {
+    String expectedUniverseDomain;
+    if (!(httpRequestInitializer instanceof HttpCredentialsAdapter)) {
+      expectedUniverseDomain = Credentials.GOOGLE_DEFAULT_UNIVERSE;
+    } else {
+      Credentials credentials = ((HttpCredentialsAdapter) httpRequestInitializer).getCredentials();
+      expectedUniverseDomain = credentials.getUniverseDomain();
+    }
+    if (!expectedUniverseDomain.equals(getUniverseDomain())) {
+      throw new IOException(
```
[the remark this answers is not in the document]
lqiu96: Ah right. Will change.

google-api-client/src/main/java/com/google/api/client/googleapis/services/AbstractGoogleClient.java
lqiu96: Sounds good, will update,

google-api-client/src/main/java/com/google/api/client/googleapis/services/AbstractGoogleClient.java
```diff
@@ -311,6 +403,33 @@ public abstract static class Builder {
...
 
+    /** User configured Universe Domain. Defaults to `googleapis.com`. */
+    String universeDomain;
+
+    /**
+     * Regex pattern to check if the URL passed in matches the default endpoint confgured from a
+     * discovery doc. Follows the format of `https://{serviceName}(.mtls).googleapis.com/`
+     */
+    Pattern defaultEndpointRegex =
+        Pattern.compile("https://[a-zA-Z]*(\\.mtls)?\\.googleapis.com/?");
```
blakeli0, line 414: I did a few quick test but I don't think this regex works, what worked for me is 
```https:\/\/[a-zA-Z]*(\.mtls)?\.googleapis.com\/?```
Can you please make sure the regex is correct and add a few tests just for the regex?

File: google-api-client/src/main/java/com/google/api/client/googleapis/services/AbstractGoogleClient.java [hunk 6]
```diff
@@ -328,9 +447,34 @@ protected Builder(
         HttpRequestInitializer httpRequestInitializer) {
       this.transport = Preconditions.checkNotNull(transport);
       this.objectParser = objectParser;
-      setRootUrl(rootUrl);
-      setServicePath(servicePath);
+      this.rootUrl = normalizeRootUrl(rootUrl);
+      this.servicePath = normalizeServicePath(servicePath);
       this.httpRequestInitializer = httpRequestInitializer;
+      this.serviceName = parseServiceName(rootUrl);
+      this.isUserConfiguredEndpoint = !defaultEndpointRegex.matcher(this.rootUrl).matches();
```
blakeli0, line 454: I'm still not convinced that we need this. Considering the following scenarios:
1. `rootUrl` is set to non-GDU, `serviceName` would be parsed to null, hence we would use `rootUrl` as is in `determiningEndpoint()`.
2. `rootUrl` is set to GDU, `isUserConfiguredEndpoint` would be false anyway, and we would re-create the `rootUrl` with `serviceName` and `universeDomain`.

lqiu96: This does work, but I think it has misleading logic inside for certain edge cases (i.e. regional endpoints). For example, when using a regional endpoint in speech: `https://eu-speech.googleapis.com` and passing that into the constructor, the logic in this PR would work. But I believe it sets the serviceName as `eu-speech` and builds the endpoint, when it _probably_ shouldn't.

The more I think about this, the more I'm leaning towards fixing the serviceName parsing logic the same way we did in GAPICs. The Apiary generator probably should be the one parsing the discovery doc and generating this value as a constant inside the library. That way we don't have worry about parsing the serviceName based on user params or user set values and the Apiary would have this value by default. I probably should have done that to begin with, but I didn't account for regional endpoints and I thought this way would be fine.

google-api-client/src/main/java/com/google/api/client/googleapis/services/AbstractGoogleClient.java
lqiu96: Oh huh, I did add unit tests for this regex and it did pass: https://github.com/googleapis/google-api-java-client/blob/1cfd93ae34f44932b135a9dafa1fa6d8698448ae/google-api-client/src/test/java/com/google/api/client/googleapis/services/AbstractGoogleClientTest.java#L328-L357.  Do you have the values you tested with? I may have tested incorrectly.

google-api-client/src/main/java/com/google/api/client/googleapis/services/AbstractGoogleClient.java
```diff
@@ -88,6 +100,74 @@ protected AbstractGoogleClient(Builder builder) {
...
+   * from the google-auth-library.
+   *
+   * <p>To use a non-GDU Credentials, you must use the HttpCredentialsAdapter class.
+   *
+   * @throws IOException if there is an error reading the Universe Domain from the credentials
+   * @throws IllegalStateException if the configured Universe Domain does not match the Universe
+   *     Domain in the Credentials
+   */
+  public void validateUniverseDomain() throws IOException {
+    if (httpRequestInitializer instanceof HttpCredentialsAdapter) {
```
blakeli0, line 157: nit: Change this to 
```
if (!(httpRequestInitializer instanceof HttpCredentialsAdapter)) {
    return;
}
```
for better readability.

pom.xml
```diff
@@ -499,6 +513,7 @@
     <project.httpclient.version>4.5.14</project.httpclient.version>
     <project.commons-codec.version>1.16.0</project.commons-codec.version>
     <project.oauth.version>1.35.0</project.oauth.version>
+    <project.auth.version>1.22.0</project.auth.version>
```
blakeli0, line 516: We might discussed this in the past, but I'm a little surprised that this project is not dependent on auth library yet. We probably can not avoid it because we need to access the `Credentials` object for validation, but let's be more careful here as it may introduce dependency problem for customers. @suztomo Do you have any concerns regarding this new auth dependency?

suztomo: This does not create a circular dependency. Looks good to me.

image

blakeli0 (approved):

PR #2459, opened 2024-03-25
chore: update DocFX profile and remove obsolete kokoro configs

### LINKED ISSUE #2458: External documentation is outdated - latest published version is 2.1.2
The last version of this library that's been published to c.g.c. is [v2.1.2](https://cloud.google.com/java/docs/reference/google-api-client/latest/overview). I double-checked the Cloud RAD bucket, and confirmed that v2.1.2 is the latest tarball that's been uploaded to the bucket.

This means there's something wrong with the release process for this library such that the tarball doesn't get generated/uploaded to the Cloud RAD bucket. Based on Fusion/ it looks like other parts of the release have been failing as well.

alicejli:
We have moved the release jobs to Google3. This removes the obsolete kokoro configs.

Javadoc generation has been broken for a while. This updates the DocFx profile such that `mvn javadoc:aggregate` can be run successfully. The other part of this fix will be within g3.

Fixes #2458☕️


pom.xml
```diff
@@ -626,35 +626,54 @@
...
+                     if [ ! -d "play-services" ]; then
+                       mkdir play-services &&
+                       cd play-services   &&
+                       curl --output play-services-basement-8.3.0.aar https://dl.google.com/dl/android/maven2/com/google/android/gms/play-services-basement/8.3.0/play-services-basement-8.3.0.aar &&
+                       unzip play-services-basement-8.3.0.aar &&
+                       mvn install:install-file -Dfile=classes.jar -DgroupId=com.google.android.google-play-services -DartifactId=google-play-services -Dversion=1 -Dpackaging=jar
+                     fi
+                    ]]></argument>
+                  </arguments>
+                </configuration>
```
burkedavison, line 657: What is this for?

suztomo: I remember there's a very strange artifact that is not published to Maven Central but available in the dl.google.com site.

lqiu96: IIRC, it's for the one-time step up to install google-play-services (meant to be done to build locally so we can run build locally): https://github.com/googleapis/google-api-java-client?tab=readme-ov-file#one-time-setup

I don't think it'll break anything, but I'm not sure if we should/need to add it in to the pom. Maybe we can put this behind a profile?

suztomo (approved):

pom.xml
alicejli: We need it for the javadoc job to run successfully; this addition is part of the `DocFx` profile

PR #2462, opened 2024-05-06
feat: Add ApiVersion Support

lqiu96:
Changes:
- Pass `X-Google-Api-Version` as a header as part of requests if apiVersion exists in the Discovery Doc

Corresponding generator change https://github.com/googleapis/google-api-java-client-services/pull/20930

This PR will go in first then merge in the generator change once this is released and in maven central.

Logs:
```
2024-05-09 14:37:57,833 CONFIG   com.google.api.client.http.HttpTransport - -------------- REQUEST  --------------
GET {URL}
Accept-Encoding: gzip
Authorization: Bearer {BEARER_TOKEN}
User-Agent: DeltaV Example Java App Google-API-Java-Client/2.4.1-SNAPSHOT Google-HTTP-Java-Client/1.44.1 (gzip)
x-goog-user-project: {PROJECT_ID}
x-goog-api-client: gl-java/17.0.10 gdcl/2.4.1 linux/6.6.15
x-google-api-version: v1_20230601
```

### DIFF
--- a/google-api-client/src/main/java/com/google/api/client/googleapis/services/AbstractGoogleClientRequest.java
+++ b/google-api-client/src/main/java/com/google/api/client/googleapis/services/AbstractGoogleClientRequest.java
@@ -64,6 +64,11 @@ public abstract class AbstractGoogleClientRequest<T> extends GenericData {
   public static final String USER_AGENT_SUFFIX = "Google-API-Java-Client";
 
   private static final String API_CLIENT_HEADER = "X-Goog-Api-Client";
+
+  /**
+   * The generated request class will pass this constant as part of the header if the RPC supports
+   * ApiVersion.
+   */
   protected static final String API_VERSION_HEADER = "X-Google-Api-Version";
 
   /** Google client. */
--- a/google-api-client/src/main/java/com/google/api/client/googleapis/services/AbstractGoogleClientRequest.java
+++ b/google-api-client/src/main/java/com/google/api/client/googleapis/services/AbstractGoogleClientRequest.java
@@ -37,7 +37,9 @@ import com.google.api.client.http.HttpResponseInterceptor;
 import com.google.api.client.http.UriTemplate;
 import com.google.api.client.util.GenericData;
 import com.google.api.client.util.Preconditions;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Joiner;
+import com.google.common.base.Strings;
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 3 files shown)


google-api-client/src/main/java/com/google/api/client/googleapis/services/AbstractGoogleClientRequest.java
```diff
@@ -62,7 +63,8 @@ public abstract class AbstractGoogleClientRequest<T> extends GenericData {
    */
   public static final String USER_AGENT_SUFFIX = "Google-API-Java-Client";
 
-  private static final String API_VERSION_HEADER = "X-Goog-Api-Client";
+  private static final String API_CLIENT_HEADER = "X-Goog-Api-Client";
+  protected static final String API_VERSION_HEADER = "X-Google-Api-Version";
```
lqiu96, line 67: protected access as this will be used in the generated client classes.

google-api-client/src/main/java/com/google/api/client/googleapis/testing/services/MockGoogleClientRequest.java
```diff
@@ -46,7 +47,32 @@ public MockGoogleClientRequest(
...
+      String method,
+      String uriTemplate,
+      HttpContent content,
+      Class<T> responseClass,
+      String apiVersion) {
     super(client, method, uriTemplate, content, responseClass);
+    // Matches generator code: Null or Empty String is not set to the header
+    if (!Strings.isNullOrEmpty(apiVersion)) {
+      getRequestHeaders().set(API_VERSION_HEADER, apiVersion);
+    }
```
lqiu96, line 75: Add a constructor with new ApiVersion param. This allows us to mimic the behavior where the generated clients pass the apiVersion to the AbstractGoogleClientRequest class.

google-api-client/src/main/java/com/google/api/client/googleapis/services/AbstractGoogleClientRequest.java
suztomo: Can you add that (the generated code uses this field to set the header) in the source code comment?

Also if there's any public document available for API versions in https://google.aip.dev, reference it from here.

suztomo (approved):

diegomarquezp (approved):

lqiu96:
> Can you add that (the generated code uses this field to set the header) in the source code comment? Also if there's any public document available for API versions in https://google.aip.dev/, reference it from here.

Will do, thanks!

PR #2484, opened 2024-05-24
feat: WebHookUtil classes using the jakarta namespace

### LINKED ISSUE #2260: Support jakarta.servlet naming over javax.servlet
I'd like to upgrade to newer versions of JakartaEE/JavaEE, but the package names have changed from javax.servlet to jakarta.servlet, so the Google Client SDK won't run (class not found errors).

This request is for the new Jakarta package names to be supported.

suztomo:
Fixes #2260

Internal bug b/342581012.

I copied the WebhookUtils and WebhookUtils (https://github.com/googleapis/google-api-java-client/tree/2d25655b94dc1ae93b5c3a97046daf6a35a9923b/google-api-client-servlet/src/main/java/com/google/api/client/googleapis/extensions/servlet/notifications ) replacing the javax namespace with the jakarta.servlet namespace.
This is the same methology as https://github.com/googleapis/google-oauth-java-client/pull/1115.

- Replaced javax with jakarta and declared the dependency with
  provided scope.
- Keeping the classes as `Beta` annotated.
- WebhookHeaders in the original package does not reference the javax namespace.
- No tests exist for the classes.
- Added `@since 2.6.0` as the last release was 2.5.1.

### DIFF
--- a/google-api-client-servlet/pom.xml
+++ b/google-api-client-servlet/pom.xml
@@ -84,6 +84,12 @@
     <dependency>
       <groupId>javax.servlet</groupId>
       <artifactId>servlet-api</artifactId>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>jakarta.servlet</groupId>
+      <artifactId>jakarta.servlet-api</artifactId>
+      <scope>provided</scope>
     </dependency>
   </dependencies>
 </project>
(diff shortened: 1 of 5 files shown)


diegomarquezp (approved):
TLDR - Jakarta webservers don't contain `javax` classes, producing `NoClassFoundError` if the user code references any of the webhooks ([example](https://developers.google.com/api-client-library/java/google-api-java-client/oauth2#web_server_applications)). This PR adds copies of the webhooks using a special package suffix `jakarta`

PR #2551, opened 2025-01-17
fix: Add warnings to users about using credentials from external sources

zhumin8:
Context: b/389124064


sai-sunder-s (approved):

sai-sunder-s:
Please do not merge.

PR #2565, opened 2025-03-13
ci: Airlock-friendly plugin dependency declarations

suztomo:
Some Maven plugins are not needed during releases. Let's use
profiles to avoid requiring the plugins unnecessarily.

### DIFF
--- a/google-api-client-appengine/pom.xml
+++ b/google-api-client-appengine/pom.xml
@@ -41,7 +41,6 @@
       <plugin>
         <groupId>org.apache.felix</groupId>
         <artifactId>maven-bundle-plugin</artifactId>
-        <version>5.1.9</version>
         <executions>
           <execution>
             <id>bundle-manifest</id>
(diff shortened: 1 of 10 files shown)


google-api-client-appengine/pom.xml
```diff
@@ -128,4 +115,29 @@
       <scope>test</scope>
     </dependency>
   </dependencies>
+  <profiles>
+    <profile>
+      <id>animal-sniffer</id>
```
meltsufin, line 120: Isn't this already declared in the parent pom?

pom.xml
```diff
@@ -589,6 +598,69 @@
       </build>
     </profile>
 
+    <profile>
+      <id>clirr-compatibility-check</id>
```
meltsufin, line 602: Would it make sense to combine `clirr-compatibility-check` and `animal-sniffer` under one profile like `compatibility-checks`?

